Issue #4, opened 2021-06-08
Create SonarLint rules configuration service


duncanp-sonar:
### Description

Add a new OmniSharp service endpoint that accepts configuration (rule severities and rule parameters) for  SonarC# rules

- [x] add service endpoint
- [x] add internal service that stores the current configuration

PR #25, opened 2021-06-15
Add RuleDefinitionsRepository

rita-sonarsource:
Part of #4


omnisharp-dotnet/src/SonarLint.OmniSharp.Plugin/packages.lock.json
```diff
@@ -11,6 +11,18 @@
           "Microsoft.NETCore.Platforms": "1.1.0"
         }
       },
+      "Newtonsoft.Json": {
```
rita-sonarsource, line 14: I uploaded Newtonsoft.Json and System.ComponentModel.Composition to the nuget feed, to use their netstandard 2.0 versions.
I also uploaded `Microsoft.NETCore.Platforms.1.1.0` -- it was a transitive dependency of `NETStandard.Library` and the build was failing without it.

omnisharp-dotnet/src/SonarLint.OmniSharp.Plugin.UnitTests/TestingInfrastructure/MefTestHelpers.cs
```diff
@@ -0,0 +1,140 @@
...
+using System.Collections.Generic;
+using System.ComponentModel.Composition;
+using System.ComponentModel.Composition.Hosting;
+using System.ComponentModel.Composition.Primitives;
+using System.Linq;
+using FluentAssertions;
+
+namespace SonarLint.VisualStudio.Integration.UnitTests
+{
+    public static class MefTestHelpers
```
rita-sonarsource, line 31: Copied as-is from SLVS, along with the other 2 files (Exceptions + SingleObjectImporter), so we could use the Mef tests helper.

omnisharp-dotnet/src/SonarLint.OmniSharp.Plugin.UnitTests/packages.lock.json
```diff
@@ -54,29 +54,6 @@
         "resolved": "16.6.1",
         "contentHash": "nBYXDgAZCfjsOVzlhMB5olGvX4dTDWB/gWaYS/MhgXBcCz8XJuVGqkfK8LmwlBR/eeUPE9Q/NFZNwlJyMZf0vg=="
       },
-      "Microsoft.CSharp": {
```
rita-sonarsource, line 57: I'm not sure where this dependency came from. The lock file was updated automatically.

rita-sonarsource:
@duncanp-sonar please review, this is an initial commit.

omnisharp-dotnet/src/SonarLint.OmniSharp.Plugin.UnitTests/packages.lock.json
duncanp-sonar: Probably from NewtonSoft - [see here](https://github.com/JamesNK/Newtonsoft.Json/issues/1270). If I remember correctly, some C# syntactic sugar keywords will cause the compiler to add a ref to that assembly (e.g. _dynamic_?)

omnisharp-dotnet/src/SonarLint.OmniSharp.Plugin/SonarLint.OmniSharp.Plugin.csproj
```diff
@@ -2,6 +2,12 @@
 
   <PropertyGroup>
     <TargetFramework>netstandard2.0</TargetFramework>
+    <LangVersion>8</LangVersion>
```
duncanp-sonar, line 5: `netstandard2.0` defaults to `v7.3`. Why would we risk targeting a higher-level version of the language?

duncanp-sonar (commented):
Looks ok, except for the question about setting the language version in the product code.

omnisharp-dotnet/src/SonarLint.OmniSharp.Plugin/SonarLint.OmniSharp.Plugin.csproj
rita-sonarsource: As we discussed offline - just for synthetic sugar "using" statement.

PR #26, opened 2021-06-15
[2] Add RulesToReportDiagnosticsConverter

rita-sonarsource:
Part of #4


omnisharp-dotnet/src/SonarLint.OmniSharp.Plugin/packages.lock.json
```diff
@@ -2,6 +2,21 @@
   "version": 1,
   "dependencies": {
     ".NETStandard,Version=v2.0": {
+      "Microsoft.CodeAnalysis.Common": {
```
rita-sonarsource, line 5: I had to add `Microsoft.CodeAnalysis.Common.3.9.0` to the nuget feed to use the netstandard 2.0 version. Unfortunately I also had to add all the transitive dependencies as well...

omnisharp-dotnet/azure-pipelines-dotnet.yml
```diff
@@ -6,6 +6,8 @@ trigger:
 
 pr:
```
rita-sonarsource, line 7: Temp fix: triggering the build pipeline on PRs that are merged to other feature branches.

omnisharp-dotnet/src/SonarLint.OmniSharp.Plugin/packages.lock.json
duncanp-sonar: This will affect the minimum version of OmniSharp that our plugin can be used with (so it would be worth adding a comment to that effect next to the package reference in the project file).

Which version is OmniSharp targetting?

From looking at NuGet, it looks like [v3+](https://www.nuget.org/packages/Microsoft.CodeAnalysis.Common/3.0.0) and later are compatible with netstandard 2.0.

omnisharp-dotnet/src/SonarLint.OmniSharp.Plugin/Rules/RulesToReportDiagnosticsConverter.cs
```diff
@@ -0,0 +1,45 @@
...
+using Microsoft.CodeAnalysis;
+
+namespace SonarLint.OmniSharp.Plugin.Rules
+{
+    internal interface IRulesToReportDiagnosticsConverter
+    {
+        Dictionary<string, ReportDiagnostic> Convert(IEnumerable<RuleDefinition> rules);
+    }
+
+    internal class RulesToRulesToReportDiagnosticsConverter : IRulesToReportDiagnosticsConverter
```
duncanp-sonar, line 32: Repeated `RulesTo` in the class name.

omnisharp-dotnet/src/SonarLint.OmniSharp.Plugin/Rules/RulesToReportDiagnosticsConverter.cs
```diff
@@ -0,0 +1,45 @@
...
+    {
+        public Dictionary<string, ReportDiagnostic> Convert(IEnumerable<RuleDefinition> rules)
+        {
+            var diagnosticOptions = rules
+                .ToDictionary(x => x.RuleId,
+                    rule => rule.IsEnabled
+                        ? ReportDiagnostic.Warn
```
duncanp-sonar, line 39: Worth adding a comment that the reported severity is handled on the Java side, and we're using `Warn` just to make sure the rule is run.

duncanp-sonar (approved):
Looks good, just the point about the minimum Microsoft.XXX package version.

Issue #28, opened 2021-06-16
Create Sonar analysis rules provider


duncanp-sonar:
### Description

Create a service that returns the complete list of Sonar analysis rules that are available.

The service will discover and load the set of assemblies containing the Sonar Roslyn rules. It then need to reflect over types in the assemblies, returning the ones that inherit from `DiagnosticAnalyzer`.

The set of rules shipped in a particular version of the extension is fixed i.e. it will not change between invocations.

TODO
Are there any analyzers we don't want to execute, such as the "utility" analyzers?
The utility analyzers don't create diagnostics; instead, the collect metadata such as lines of code, info about location of keywords for syntax highlighting in SonarQube/SonarCloud etc. If I remember correctly, the utility analyzers detect whether they are being run by the Scanner for .NET and will no-op otherwise i.e. we might not need to do anything.

#### Location of the analyzer assemblies
By convention, the Java plugin will unpack the SonarC# analyzer assemblies and any dependencies into a sub-folder called `analyzers`. The sub-folder will be a direct child of the folder that contains our .NET OmniSharp extension assembly.

PR #34, opened 2021-06-16
[1] Add SonarLintHostServicesProvider

rita-sonarsource:
Part of #28

### DIFF
--- a/omnisharp-dotnet/Directory.Build.props
+++ b/omnisharp-dotnet/Directory.Build.props
@@ -15,6 +15,7 @@
 
     <OmniSharpLocalDownloadDir>$(MSBuildThisFileDirectory).omnisharp\$(OmniSharpVersion)</OmniSharpLocalDownloadDir>
     <OmniSharpUnzippedDir>$(OmniSharpLocalDownloadDir)\unzipped</OmniSharpUnzippedDir>
+    
   </PropertyGroup>
   <Import Project="build\DownloadOmniSharp.targets"/>
 
@@ -24,5 +25,29 @@
     <Reference Include="$(OmniSharpUnzippedDir)\omnisharp\OmniSharp.Roslyn.dll" />
     <Reference Include="$(OmniSharpUnzippedDir)\omnisharp\OmniSharp.Roslyn.CSharp.dll" />
   </ItemGroup>
+
+  <!-- ****************************************************-->
+  <!-- sonar-dotnet assembly references                       -->
+  <!-- ****************************************************-->
+  <PropertyGroup>
+    <SonarDotnetAnalyzerVersion>8.24.0.32949</SonarDotnetAnalyzerVersion>
+  </PropertyGroup>
+  
+  <ItemGroup Label="SonarDotnetAnalyzers">
+    <PackageReference Include="SonarAnalyzer.CSharp" 
+                      Version="$(SonarDotnetAnalyzerVersion)" 
+                      IncludeAssets="none" 
+                      ExcludeAssets="all" 
+                      GeneratePathProperty="true" />
+  </ItemGroup>
+
+  <Target Name="IncludeSonarAnalyzersInOutput" BeforeTargets="BeforeBuild">
+    <Message Importance="high" Text="XXX Locating SonarAnalyzer assemblies to copy: $(PkgSonarAnalyzer_CSharp)" />
+    <ItemGroup>
+      <AnalyzerFiles Include="$(PkgSonarAnalyzer_CSharp)\**\analyzers\*.*"/>
+    </ItemGroup>
+    <Copy SourceFiles="@(AnalyzerFiles)" 
+          DestinationFolder="$(OutputPath)analyzers\" />
+  </Target>
 
 </Project>
--- a/omnisharp-dotnet/src/SonarLint.OmniSharp.Plugin.UnitTests/DiagnosticWorker/SonarLintHostServicesProviderTests.cs
+++ b/omnisharp-dotnet/src/SonarLint.OmniSharp.Plugin.UnitTests/DiagnosticWorker/SonarLintHostServicesProviderTests.cs
@@ -1,0 +1,78 @@
+﻿/*
+ * SonarOmnisharp
+ * Copyright (C) 2021-2021 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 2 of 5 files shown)


omnisharp-dotnet/src/SonarLint.OmniSharp.Plugin.UnitTests/DiagnosticWorker/SonarLintHostServicesProviderTests.cs
```diff
@@ -0,0 +1,78 @@
...
+        public void MefCtor_CheckIsExported()
+        {
+            MefTestHelpers.CheckTypeCanBeImported<SonarLintHostServicesProvider, ISonarLintHostServicesProvider>(null, new []
+            {
+                MefTestHelpers.CreateExport<IAssemblyLoader>(Mock.Of<IAssemblyLoader>())
+            });
+        }
+
+        [TestMethod]
+        public void Assemblies_LoadsSonarDotnetAssemblies()
```
rita-sonarsource, line 44: This is effectively an integration test for the directory.build.props logic

omnisharp-dotnet/src/SonarLint.OmniSharp.Plugin/DiagnosticWorker/SonarLintHostServicesProvider.cs
```diff
@@ -0,0 +1,56 @@
...
+    [PartCreationPolicy(CreationPolicy.Shared)]
+    internal class SonarLintHostServicesProvider : ISonarLintHostServicesProvider
+    {
+        public ImmutableArray<Assembly> Assemblies { get; }
+
+        [ImportingConstructor]
+        public SonarLintHostServicesProvider(IAssemblyLoader loader)
+        {
+            var builder = ImmutableArray.CreateBuilder<Assembly>();
+            var analyzersDirectory = Path.Combine(Path.GetDirectoryName(GetType().Assembly.Location), "analyzers");
```
rita-sonarsource, line 44: As we discussed previously, this is based on convention that the analyzers will be in the output directory, in "analyzers" subfolder. 
I originally placed them next to `.omnisharp`, but it's harder to access it then from the code. Let me know if you have another idea, I'm not so keen on throwing it in the output directory.

omnisharp-dotnet/Directory.Build.props
```diff
@@ -25,4 +26,28 @@
...
 
+  <!-- ****************************************************-->
+  <!-- sonar-dotnet assembly references                       -->
+  <!-- ****************************************************-->
+  <PropertyGroup>
+    <SonarDotnetAnalyzerVersion>8.23.0.32424</SonarDotnetAnalyzerVersion>
+  </PropertyGroup>
+  
+  <ItemGroup Label="SonarDotnetAnalyzers">
+    <PackageReference Include="SonarAnalyzer.CSharp"
```
rita-sonarsource, line 37: At the moment I'm adding the nuget package out of convenience to get the `PkgSonarAnalyzer_CSharp` path, like in the prototype. I suppose we would have to change this task in the future: have the java side throw the dlls somewhere on disk (and potentially then copy them to the output directory)

duncanp-sonar:
FYI there's a separate ticket for this: https://github.com/SonarSource/sonarlint-omnisharp/issues/28

omnisharp-dotnet/src/SonarLint.OmniSharp.Plugin/DiagnosticWorker/SonarLintHostServicesProvider.cs
```diff
@@ -0,0 +1,56 @@
...
+using System.Collections.Immutable;
+using System.ComponentModel.Composition;
+using System.Diagnostics;
+using System.IO;
+using System.Reflection;
+using OmniSharp.Services;
+
+namespace SonarLint.OmniSharp.Plugin.DiagnosticWorker
+{
+    internal interface ISonarLintHostServicesProvider : IHostServicesProvider
```
duncanp-sonar, line 30: This name doesn't really describe what the service does. `IAnalyserAssemblyProvider` would be more accurate.

omnisharp-dotnet/src/SonarLint.OmniSharp.Plugin/DiagnosticWorker/SonarLintHostServicesProvider.cs
duncanp-sonar: I don't understand the question. The Java plugin is responsible for unpacking our assembly and the analyzer assemblies and laying them out on disc. It will place the analyzer assemblies in a folder called `analyzers` under our plugin, so that's where we have to look for them.

omnisharp-dotnet/Directory.Build.props
duncanp-sonar: Having any analyzer assemblies on disk is just a debugging convenience for us when building and testing just our plugin end to end. They don't even have to be the SonarC# assemblies.  We don't need them for unit testing.

omnisharp-dotnet/src/SonarLint.OmniSharp.Plugin/DiagnosticWorker/SonarLintHostServicesProvider.cs
```diff
@@ -0,0 +1,56 @@
...
+            var analyzersDirectory = Path.Combine(Path.GetDirectoryName(GetType().Assembly.Location), "analyzers");
+
+            foreach (var filePath in Directory.GetFiles(analyzersDirectory))
+            {
+                var fullPath = Path.Combine(analyzersDirectory, filePath);
+                Debug.Assert(File.Exists(fullPath), $"Analyzer assembly could not be found: {fullPath}");
+                builder.Add(loader.LoadFrom(fullPath));
+            }
+
+            Assemblies = builder.ToImmutable();
```
duncanp-sonar, line 53: When I discussed this with Julien we agreed we should throw an exception if we couldn't find any assemblies (it would indicate a problem on the Java side in how it is unpacking and configuring OmniSharp - throwing an exception would make it easier for them to detect/test. It shouldn't happen at runtime).

omnisharp-dotnet/Directory.Build.props
duncanp-sonar: I think there should be an explanation in the project file to make it clear that the analyzer assemblies are being added here just as a debugging convenience when working in VS.

Having a variable for `SonarDotNetAnalyzerVersion` gives the impression that it matters which version is used, which it doesn't.

omnisharp-dotnet/src/SonarLint.OmniSharp.Plugin.UnitTests/DiagnosticWorker/SonarLintHostServicesProviderTests.cs
duncanp-sonar: This is a bit tricky.  From a product point of view, we don't need an _integration_ test that our project correctly dumps the Sonar analyzers in the correct folder, since that isn't part of our responsibility. 

There needs to be a _unit_ test that the services locates and loads any assemblies that are in the correct folder. However, the product code would need to use an `IFileSystem` object so the set of assemblies could be correctly mocked.

PR #35, opened 2021-06-16
[2] Add SonarLintCodeActionProvider

rita-sonarsource:
Part of #28


omnisharp-dotnet/src/SonarLint.OmniSharp.Plugin/DiagnosticWorker/SonarLintCodeActionProvider.cs
```diff
@@ -0,0 +1,40 @@
...
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System.ComponentModel.Composition;
+using OmniSharp.Services;
+
+namespace SonarLint.OmniSharp.Plugin.DiagnosticWorker
+{
+    internal interface ISonarLintCodeActionProvider : ICodeActionProvider
```
duncanp-sonar, line 26: Some design comments would be useful e.g. 

> We only want this provider to be used by our custom diagnostic worker and not the "normal" OmniSharp workers, so we're exporting it using a different inteface.
However, we're reusing the OmniSharp `AbstractCodeActionProvider` class because it makes it easy to load assemblies from and extract diagnostic analyzers from them.

duncanp-sonar (approved):
Just needs a comment about the design.

PR #36, opened 2021-06-17
Add SonarAnalyzerAssembliesProvider

### LINKED ISSUE #28: Create Sonar analysis rules provider

rita-sonarsource:
Fixes #28


omnisharp-dotnet/src/SonarLint.OmniSharp.Plugin/DiagnosticWorker/SonarAnalyzerAssembliesProvider.cs
```diff
@@ -30,4 +34,57 @@ internal interface ISonarAnalyzerAssembliesProvider
...
+        internal static string AnalyzersDirectory { get; } =
+            Path.Combine(Path.GetDirectoryName(typeof(SonarAnalyzerAssembliesProvider).Assembly.Location), "analyzers");
+        
+        private readonly IAssemblyLoader loader;
+        private readonly Func<string, string[]> getFilesInDirectory;
+        private ImmutableArray<Assembly> loadedAssemblies;
+
+        public ImmutableArray<Assembly> Assemblies
+        {
+            get
```
rita-sonarsource, line 54: I decided to make it lazy, as otherwise the MEF test was failing due to missing directory, and I didn't want to create it.

omnisharp-dotnet/src/SonarLint.OmniSharp.Plugin/DiagnosticWorker/SonarAnalyzerAssembliesProvider.cs
```diff
@@ -30,4 +34,57 @@ internal interface ISonarAnalyzerAssembliesProvider
...
+                    loadedAssemblies = LoadAssemblies();
+                }
+
+                return loadedAssemblies;
+            }
+        }
+
+        [ImportingConstructor]
+        public SonarAnalyzerAssembliesProvider(IAssemblyLoader loader)
+            : this(loader, Directory.GetFiles)
```
rita-sonarsource, line 67: Since we only need this one method, I decided not to add a dependency on System.IO.Abstractions for now.

omnisharp-dotnet/src/SonarLint.OmniSharp.Plugin/DiagnosticWorker/SonarAnalyzerAssembliesProvider.cs
```diff
@@ -30,4 +34,57 @@ internal interface ISonarAnalyzerAssembliesProvider
...
+            this.getFilesInDirectory = getFilesInDirectory;
+        }
+
+        private ImmutableArray<Assembly> LoadAssemblies()
+        {
+            var builder = ImmutableArray.CreateBuilder<Assembly>();
+
+            foreach (var filePath in getFilesInDirectory(AnalyzersDirectory))
+            {
+                var analyzerAssembly = loader.LoadFrom(filePath);
```
rita-sonarsource, line 83: I'm not sure what to do with exception handling.
Omnisharp's assembly loader will catch all exceptions and just return a null.
https://github.com/OmniSharp/omnisharp-roslyn/blob/master/src/OmniSharp.Host/Services/AssemblyLoader.cs#L85

I'm not sure what our behavior should be: throw an exception and let it propagate to our worker and our endpoint?
Perhaps we should revisit it later when we have all the pieces in place.

rita-sonarsource:
@duncanp-sonar please review. I closed the previous PR and opened this one, that doesn't have any SonarCSharp packages. I'll add that in the future because I don't need it for now.

omnisharp-dotnet/src/SonarLint.OmniSharp.Plugin/DiagnosticWorker/SonarAnalyzerAssembliesProvider.cs
duncanp-sonar: > throw an exception and let it propagate to our worker and our endpoint?

We want the failure to be obvious to the Java code during development.

I don't think we need to do anything special if the directory doesn't exist if that already throws an exception that will be propagated through the service.

I think we should throw if we don't find any assemblies; if we don't, it will be harder to track down from the Java.

duncanp-sonar: Also: let's throw on null assemblies too. We don't think it should happen if the plugin packaging is correct. However, there might be cross-platform cases where that we haven't though of/tested that will cause failures in practice. In that case, failing fast would help us and users.

duncanp-sonar (commented):
Looks good. See my comments on error handling: I think we should fail fast and throw. What do you think?

duncanp-sonar:
@rita-sonarsource FYI I'll pick this up and make the changes to throw on any error.

PR #38, opened 2021-06-17
Pipeline: publish jar content files as a pipeline artifact

duncanp-sonar:
Amended build and pipeline to publish the files that need to be included in the jar to a folder called `JarContentFiles`.

The pipeline publishes the contents of the folder as an artifact that can be downloaded after the pipeline has finished.

### DIFF
--- a/omnisharp-dotnet/src/SonarLint.OmniSharp.Plugin/SonarLint.OmniSharp.Plugin.csproj
+++ b/omnisharp-dotnet/src/SonarLint.OmniSharp.Plugin/SonarLint.OmniSharp.Plugin.csproj
@@ -8,7 +8,9 @@
     <!-- We need to target the same Roslyn version as is used by OmniSharp --> 
     <PackageReference Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="3.10.0" />
 
-    <PackageReference Include="Newtonsoft.Json" Version="13.0.1" />
+    <PackageReference Include="Newtonsoft.Json" Version="13.0.1">
+      <GeneratePathProperty>True</GeneratePathProperty>
+    </PackageReference>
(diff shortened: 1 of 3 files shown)


omnisharp-dotnet/src/SonarLint.OmniSharp.Plugin/SonarLint.OmniSharp.Plugin.csproj
```diff
@@ -11,5 +11,27 @@
     <PackageReference Include="Newtonsoft.Json" Version="13.0.1" />
     <PackageReference Include="System.ComponentModel.Composition" Version="5.0.0" />
   </ItemGroup>
+  
+  
+  <!-- ********************************************** -->
+  <!-- Packaging files to be included in the Java jar -->
```
duncanp-sonar, line 17: I looked at using the `dotnet publish` command, but couldn't work out to exclude a lot of files don't want to include. I'm hopefully we might be able to use it in the future, but for now I just added a build target.

omnisharp-dotnet/src/SonarLint.OmniSharp.Plugin/SonarLint.OmniSharp.Plugin.csproj
```diff
@@ -11,5 +11,27 @@
...
+  <!-- ********************************************** -->
+    <!-- Add any files that should be included in the Java plugin to the item group @"FilesToIncludeInJar".
+         The files will be copied to the directory "JarContentFiles". -->
+  <ItemGroup>
+    <FilesToIncludeInJar Include="$(TargetDir)$(TargetFileName)" />
```
duncanp-sonar, line 22: Currently I'm only publishing the plugin dll itself but there will be others e.g. the NewtonSoft dll (which for some reason isn't being copied to the `bin` folder as part of the build, but that's a separate issue).

PR #50, opened 2021-06-20
Add SonarLintDiagnosticWorker

rita-sonarsource:
Part of #5


omnisharp-dotnet/src/SonarLint.OmniSharp.Plugin/DiagnosticWorker/OmniSharp/Configuration.cs
```diff
@@ -0,0 +1,11 @@
+﻿namespace OmniSharp
+{
+    /// <summary>
+    /// Copied unchanged from OmniSharp.Configuration
```
rita-sonarsource, line 4: Not sure if "unchanged" is the right word here; I didn't copy the entire class, just the one const that was needed.

omnisharp-dotnet/src/SonarLint.OmniSharp.Plugin/DiagnosticWorker/OmniSharp/DiagnosticExtensions.cs
```diff
@@ -0,0 +1,57 @@
...
+using OmniSharp.Models.Diagnostics;
+using OmniSharp.Roslyn.CSharp.Services.Diagnostics;
+using System.Collections.Generic;
+using System.Collections.Immutable;
+using System.Linq;
+
+namespace OmniSharp.Helpers
+{
+    /// <summary>
+    /// Copied unchanged from OmniSharp.Helpers.DiagnosticExtensions
```
rita-sonarsource, line 11: This is "unchanged" for now; it will change when we add secondary locations

omnisharp-dotnet/src/SonarLint.OmniSharp.Plugin/DiagnosticWorker/OmniSharp/CopiedCSharpDiagnosticWorkerWithAnalyzers.cs
```diff
@@ -0,0 +1,374 @@
...
+using OmniSharp.Options;
+using OmniSharp.Roslyn.CSharp.Workers.Diagnostics;
+using OmniSharp.Services;
+
+namespace OmniSharp.Roslyn.CSharp.Services.Diagnostics
+{
+    /// <summary>
+    /// Copied from OmniSharp.Roslyn.CSharp.Services.Diagnostics.CSharpDiagnosticWorkerWithAnalyzers
+    /// https://github.com/OmniSharp/omnisharp-roslyn/blob/master/src/OmniSharp.Roslyn.CSharp/Workers/Diagnostics/CSharpDiagnosticWorkerWithAnalyzers.cs
+    /// Nothing changed except making <see cref="AnalyzeDocument"/> protected-virtual
```
rita-sonarsource, line 26: I noticed in the prototype files that eventually all the modifications lead to AnalyzeDocument method -- this was very fortunate because it meant that we could minimize the changes to making this method `protected virtual` and overriding it.
This way we can diff the entire file easily in future omnisharp versions.

omnisharp-dotnet/src/SonarLint.OmniSharp.Plugin/DiagnosticWorker/SonarLintDiagnosticWorker.cs
```diff
@@ -0,0 +1,71 @@
...
+            : base(workspace, new[] {sonarAnalyzerCodeActionProvider}, loggerFactory, forwarder, options.CurrentValue)
+        {
+            this.sonarLintDiagnosticWorkerDataProvider = sonarLintDiagnosticWorkerDataProvider;
+        }
+        
+        protected override Task<ImmutableArray<Diagnostic>> AnalyzeDocument(Project project,
+            ImmutableArray<DiagnosticAnalyzer> allAnalyzers, Compilation compilation,
+            AnalyzerOptions workspaceAnalyzerOptions, Document document)
+        {
+            var diagnosticWorkerData = sonarLintDiagnosticWorkerDataProvider.Get(compilation, workspaceAnalyzerOptions);
```
rita-sonarsource, line 62: All our modifications are now scoped to this single method and written in a separate class.

omnisharp-dotnet/src/SonarLint.OmniSharp.Plugin/DiagnosticWorker/SonarLintDiagnosticWorker.cs
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    [Export(typeof(ISonarLintDiagnosticWorker))]
+    [PartCreationPolicy(CreationPolicy.Shared)]
+    internal class SonarLintDiagnosticWorker : CopiedCSharpDiagnosticWorkerWithAnalyzers, ISonarLintDiagnosticWorker
+    {
+        private readonly ISonarLintDiagnosticWorkerDataProvider sonarLintDiagnosticWorkerDataProvider;
+
+        [ImportingConstructor]
+        public SonarLintDiagnosticWorker(ISonarLintDiagnosticWorkerDataProvider sonarLintDiagnosticWorkerDataProvider,
```
rita-sonarsource, line 47: I took the parameters based on the `SonarLintCSharpDiagnosticWorkerFactory` class from the prototype. I didn't see the need to create a factory, not sure if it's needed: it was used by `SonarLintCodeCheckService` that created a worker in the ctor. The base `CopiedCSharpDiagnosticWorkerWithAnalyzers` has a dictionary so I believe this should be a singleton as well.

omnisharp-dotnet/src/SonarLint.OmniSharp.Plugin/DiagnosticWorker/SonarLintDiagnosticWorkerDataProvider.cs
```diff
@@ -0,0 +1,133 @@
...
+namespace SonarLint.OmniSharp.Plugin.DiagnosticWorker
+{
+    internal class DiagnosticWorkerModifications
+    {
+        public ImmutableArray<DiagnosticAnalyzer> Analyzers { get; set; }
+        public Compilation Compilation { get; set; }
+        public AnalyzerOptions AnalyzerOptions { get; set; }
+    }
+    
+    internal interface ISonarLintDiagnosticWorkerDataProvider
```
rita-sonarsource, line 40: Not sure about the name of the class and the returned result object...

omnisharp-dotnet/src/SonarLint.OmniSharp.Plugin/Rules/RulesToAdditionalTextConverter.cs
```diff
@@ -65,7 +65,7 @@ public AdditionalText Convert(IEnumerable<RuleDefinition> rules)
         }
 
         // There isn't a public implementation of source text so we need to create one
-        private class AdditionalTextImpl : AdditionalText
+        internal class AdditionalTextImpl : AdditionalText
```
rita-sonarsource, line 68: I'm using it in the tests so I changed it to internal.

omnisharp-dotnet/src/SonarLint.OmniSharp.Plugin/SonarLint.OmniSharp.Plugin.csproj
```diff
@@ -16,6 +16,8 @@
     <PackageReference Include="Newtonsoft.Json" Version="12.0.3" />
     
     <PackageReference Include="System.ComponentModel.Composition" Version="5.0.0" />
+    <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="3.1.12" />
```
rita-sonarsource, line 19: Copied OmniSharp's code dependencies

omnisharp-dotnet/src/SonarLint.OmniSharp.Plugin.UnitTests/DiagnosticWorker/SonarAnalyzerCodeActionProviderTests.cs
```diff
@@ -73,9 +73,8 @@ public void CodeDiagnosticAnalyzerProviders_LoadsFromGivenAssemblies()
         {
             var testSubject = CreateTestSubject(GetType().Assembly);
 
-            testSubject.CodeDiagnosticAnalyzerProviders.Length.Should().Be(2);
-            testSubject.CodeDiagnosticAnalyzerProviders[0].Should().BeOfType<DummyAnalyzer1>();
-            testSubject.CodeDiagnosticAnalyzerProviders[1].Should().BeOfType<DummyAnalyzer2>();
+            testSubject.CodeDiagnosticAnalyzerProviders.Should().Contain(x => x is DummyAnalyzer1);
```
rita-sonarsource, line 76: I'm creating additional analyzers for the other tests, which caused the magic number "2" to fail. I decided to instead just check that the providers array contains the 2 that I expect, ignoring the others.

omnisharp-dotnet/src/SonarLint.OmniSharp.Plugin.UnitTests/DiagnosticWorker/SonarLintDiagnosticWorkerTests.cs
```diff
@@ -0,0 +1,303 @@
...
+using OmniSharp.Roslyn;
+using OmniSharp.Services;
+using SonarLint.OmniSharp.Plugin.DiagnosticWorker;
+using SonarLint.OmniSharp.Plugin.Rules;
+using SonarLint.VisualStudio.Integration.UnitTests;
+
+namespace SonarLint.OmniSharp.Plugin.UnitTests.DiagnosticWorker
+{
+    [TestClass]
+    public class SonarLintDiagnosticWorkerTests
```
rita-sonarsource, line 48: This class has complicated setups since the `Complication` object and `OmniSharpWorkspace` need to be valid. Hence I had to create a bunch of helper methods.

I thought about extracting it to a separate helper class, but those methods shouldn't be used in any other place as they're very tightly-coupled to OmniSharp's implementation in the copied worker class. 

I tried to keep the tests themselves simple: in every test I create a valid workspace with a test file (in-memory) and run the diagnostics. I then modify our analyzers / additional files / severities and run the diagnostics again to make sure that they're picked up.

omnisharp-dotnet/src/SonarLint.OmniSharp.Plugin.UnitTests/DiagnosticWorker/SonarLintDiagnosticWorkerTests.cs
```diff
@@ -0,0 +1,303 @@
...
+            {
+                context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.Analyze | GeneratedCodeAnalysisFlags.ReportDiagnostics);
+                context.EnableConcurrentExecution();
+                context.RegisterSymbolAction(AnalyzeSymbol, SymbolKind.NamedType);
+            }
+
+            private void AnalyzeSymbol(SymbolAnalysisContext context)
+            {
+                // Hacky way to test that the analyzer options are being overriden
+                var shouldIgnoreAnalysis = context.Options.AdditionalFiles.Any(x => x.Path.Equals(Descriptor.Id));
```
rita-sonarsource, line 269: This is very ugly, used by the `AnalyzeDocument_AnalyzerOptionsAreOverriden` test: I couldn't override anything in the `AnalyzerOptions` object besides the additional files property, so I had to "signal" to the analyzer somehow that I don't want to run the diagnostics (so I could compare in the test that the changes were picked up).
I thought about inverting the condition, "if has additional file then execute, else ignore", but it meant having to add further setup to the other tests.
Let me know if you have a better idea :/

omnisharp-dotnet/src/SonarLint.OmniSharp.Plugin.UnitTests/DiagnosticWorker/SonarLintDiagnosticWorkerDataProviderTests.cs
```diff
@@ -0,0 +1,222 @@
...
+        }
+        
+        [TestMethod]
+        public void Get_HasExistingSonarAdditionalFile_OverridesSonarAdditionalFile()
+        {
+            var existingSonarAdditionalFile = new RulesToAdditionalTextConverter.AdditionalTextImpl("a/test/sonar.xml", "some content1");
+            var existingUnrelatedAdditionalFile = new RulesToAdditionalTextConverter.AdditionalTextImpl("asonar.xml", "some content2");
+            var existingOptions = CreateOptions(existingSonarAdditionalFile, existingUnrelatedAdditionalFile);
+            
+            var sonarAdditionalFile = new RulesToAdditionalTextConverter.AdditionalTextImpl("b/sonar.xml", "some new content");
```
rita-sonarsource, line 122: Originally this was `b\\SONAR.XML` to test the case sensitivity, but our build pipeline was failing. I believe this is due to this remark that on linux and max the behavior is always case-sensitive: https://docs.microsoft.com/en-us/dotnet/api/system.stringcomparison?view=net-5.0#remarks

omnisharp-dotnet/src/SonarLint.OmniSharp.Plugin/DiagnosticWorker/OmniSharp/Configuration.cs
duncanp-sonar: "Copied with other constants deleted"?

omnisharp-dotnet/src/SonarLint.OmniSharp.Plugin/DiagnosticWorker/OmniSharp/CopiedCSharpDiagnosticWorkerWithAnalyzers.cs
duncanp-sonar: Sounds good. The code will be doing additional unnecessary work to calculate some settings that we will then throw away. However, we could always change this in the future if it affects the performance.

omnisharp-dotnet/src/SonarLint.OmniSharp.Plugin/DiagnosticWorker/SonarLintDiagnosticWorker.cs
duncanp-sonar: I don't think we need to worry about lazy invocation - I think the OmniSharp host handles it for us i.e. it won't even instantiate the service handler class until the service is actually called.

omnisharp-dotnet/src/SonarLint.OmniSharp.Plugin/DiagnosticWorker/SonarLintDiagnosticWorkerDataProvider.cs
duncanp-sonar: ISonarLintAnalysisConfigProvider / ISonarLintAnalysisOptionsProvider?

omnisharp-dotnet/src/SonarLint.OmniSharp.Plugin/DiagnosticWorker/SonarLintDiagnosticWorker.cs
```diff
@@ -0,0 +1,71 @@
...
+            OmniSharpWorkspace workspace,
+            ISonarAnalyzerCodeActionProvider sonarAnalyzerCodeActionProvider,
+            ILoggerFactory loggerFactory,
+            DiagnosticEventForwarder forwarder,
+            IOptionsMonitor<OmniSharpOptions> options)
+            : base(workspace, new[] {sonarAnalyzerCodeActionProvider}, loggerFactory, forwarder, options.CurrentValue)
```
duncanp-sonar, line 53: Could we pass an empty array of code action providers to the base class?
It will still look for analyzers specified in the project, and we will still override whatever it configuration it produces anyway, so there doesn't seem to be any need to pass in our provider here.

That would mean we could drop `sonarAnalyzerCodeActionProvider` from the list of imports in this constructor.

omnisharp-dotnet/src/SonarLint.OmniSharp.Plugin.UnitTests/DiagnosticWorker/SonarAnalyzerCodeActionProviderTests.cs
duncanp-sonar: Probably worth adding that as a comment in the test.

omnisharp-dotnet/src/SonarLint.OmniSharp.Plugin.UnitTests/DiagnosticWorker/SonarLintDiagnosticWorkerDataProviderTests.cs
```diff
@@ -0,0 +1,222 @@
...
+        {
+            MefTestHelpers.CheckTypeCanBeImported<SonarLintDiagnosticWorkerDataProvider, ISonarLintDiagnosticWorkerDataProvider>(null, new []
+            {
+                MefTestHelpers.CreateExport<IRuleDefinitionsRepository>(Mock.Of<IRuleDefinitionsRepository>()),
+                MefTestHelpers.CreateExport<ISonarAnalyzerCodeActionProvider>(Mock.Of<ISonarAnalyzerCodeActionProvider>())
+            });
+        }
+
+        [TestMethod]
+        public void Get_ReturnsSonarAnalyzers()
```
duncanp-sonar, line 50: Strictly speaking this method doesn't know whether it's returning Sonar analyzers or not. The test is checking whether it returns the _expected_ analyzers.
Same for the severities test.

omnisharp-dotnet/src/SonarLint.OmniSharp.Plugin.UnitTests/DiagnosticWorker/SonarLintDiagnosticWorkerTests.cs
duncanp-sonar: Why not store the supplied additional files in a public property on this class?
The tests instantiate the class, so they can check the contents of the property after the analysis has completed.

duncanp-sonar (commented):
Looks good. A few comments inline, plus the license changes we discussed offline.

omnisharp-dotnet/src/SonarLint.OmniSharp.Plugin.UnitTests/DiagnosticWorker/SonarLintDiagnosticWorkerTests.cs
rita-sonarsource: Nice, good idea 👍

PR #51, opened 2021-06-21
Pipeline: set Sonar project version for analysis

duncanp-sonar:


omnisharp-dotnet/azure-pipelines-dotnet.yml
```diff
@@ -27,6 +27,22 @@ steps:
   inputs:
     secureFile: SonarSourceSecret.snk
 
+- powershell: |
```
duncanp-sonar, line 30: Based on the equivalent PowerShell in the SLVS pipeline.

This version number is only used for setting the analysis version currently - our assembly build number is still hardcoded.

The current value of this variable is "1.0-snapshot", so we couldn't use this value directly for the assembly version number anyway (SonarQube treats the version as a label, so it doesn't mind about the `-snapshot` suffix).

rita-sonarsource (approved):

PR #52, opened 2021-06-21
Add OmniSharp license headers

rita-sonarsource:
Part of #5

### DIFF
--- a/omnisharp-dotnet/THIRD-PARTY-NOTICES
+++ b/omnisharp-dotnet/THIRD-PARTY-NOTICES
@@ -1,0 +1,28 @@
+SonarOmnisharp uses third-party libraries or other resources
+that may be distributed under licenses different than the SonarOmnisharp
+
+Attributions and licence notices for test cases originally authored by
+third parties can be found in the respective test directories.
+
+In the event that we accidentally failed to list a required notice, please
+bring it to our attention by posting an issue in the following project:
+https://github.com/SonarSource/sonarlint-omnisharp
+
+The attached notices are provided for information only.
+
+---------------------------------------------------------------------------
+
+omnisharp-roslyn
+https://github.com/OmniSharp/omnisharp-roslyn/blob/master/license.md
+
+MIT License
+
+Copyright (c) .NET Foundation and Contributors All Rights Reserved
+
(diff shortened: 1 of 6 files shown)


omnisharp-dotnet/THIRD-PARTY-NOTICES
```diff
@@ -0,0 +1,28 @@
+SonarOmnisharp uses third-party libraries or other resources
```
rita-sonarsource, line 1: Based on SLVS https://github.com/SonarSource/sonarlint-visualstudio/blob/master/THIRD-PARTY-NOTICES

omnisharp-dotnet/src/SonarLint.OmniSharp.Plugin/SonarLint.OmniSharp.Plugin.csproj
```diff
@@ -27,7 +27,7 @@
     <!-- Add any files that should be included in the Java plugin to the item group @"FilesToIncludeInJar".
          The files will be copied to the directory "JarContentFiles". -->
   <ItemGroup>
-    <FilesToIncludeInJar Include="$(TargetDir)$(TargetFileName)" />
+    <FilesToIncludeInJar Include="$(TargetDir)$(TargetFileName);..\..\THIRD-PARTY-NOTICES;" />
```
rita-sonarsource, line 30: It doesn't seem that the java side has something like this, so no conflict (yet) -- but maybe I should check-in the file under the root repository and have the java side package it?

duncanp-sonar: I'd leave it for the time being; we can discuss with Julien when we look at the overall packaging.

`FilesToIncludeInJar` is an ItemGroup so its values can be spread across multiple elements, rather than having one element with a semi-colon separated list. The only difference is in readability.
I would normally have one file per element, but it doesn't make much difference since we only have two items at the moment.

duncanp-sonar (approved):

omnisharp-dotnet/src/SonarLint.OmniSharp.Plugin/SonarLint.OmniSharp.Plugin.csproj
rita-sonarsource: Ah right right, I changed it.

PR #53, opened 2021-06-21
Ignore compiler warnings and errors

### LINKED ISSUE #5: Create custom SonarLint diagnostic worker class
### Description

In OmniSharp, all of actual C# analyses are handled by the `CSharpDiagnosticWorkerWithAnalyzers` class.
The custom SonarLint version should work the same, except:

* any rules configuration in ruleset or .editorconfig files will be ignored.
* only SonarC# rules should be executed
* the rules to executed and rules configuration are specified by calling the SonarLint rules configuration service.
* compiler warnings and errors should not be included in the returned results.
* secondary locations should be returned, where appropriate.

- [x] create a custom "diagnostic worker" class.
- [x] check copyright requirements - ?do we need to include a copyright notice for the code that has been copied and modified?
- [x] use the SonarLint rules configuration service to create appropriate CompilerOptions, AnalyzerOptions and AdditionalFiles.
- [x] filter out compiler warnings and errors

rita-sonarsource:
Fixes #5

### DIFF
--- a/omnisharp-dotnet/src/SonarLint.OmniSharp.Plugin.UnitTests/DiagnosticWorker/SonarLintAnalysisConfigProviderTests.cs
+++ b/omnisharp-dotnet/src/SonarLint.OmniSharp.Plugin.UnitTests/DiagnosticWorker/SonarLintAnalysisConfigProviderTests.cs
@@ -42,19 +42,30 @@ namespace SonarLint.OmniSharp.Plugin.UnitTests.DiagnosticWorker
             MefTestHelpers.CheckTypeCanBeImported<SonarLintAnalysisConfigProvider, ISonarLintAnalysisConfigProvider>(null, new []
             {
                 MefTestHelpers.CreateExport<IRuleDefinitionsRepository>(Mock.Of<IRuleDefinitionsRepository>()),
-                MefTestHelpers.CreateExport<ISonarAnalyzerCodeActionProvider>(Mock.Of<ISonarAnalyzerCodeActionProvider>())
+                MefTestHelpers.CreateExport<ISonarAnalyzerCodeActionProvider>(CreateSonarCodeActionProvider(ImmutableArray<DiagnosticAnalyzer>.Empty))
             });
         }
 
         [TestMethod]
         public void Get_ReturnsCorrectAnalyzers()
         {
-            var analyzers = new DiagnosticAnalyzer[] {new DummyAnalyzer()}.ToImmutableArray();
+            var analyzers = new DiagnosticAnalyzer[] {new DummyAnalyzer()};
 
             var testSubject = CreateTestSubject(analyzers: analyzers);
             var analysisConfig = testSubject.Get(CreateCompilation(), CreateOptions());
 
             analysisConfig.Analyzers.Should().BeEquivalentTo(analyzers);
+        }
+        
+        [TestMethod]
+        public void Get_ReturnsCorrectAnalyzerRules()
+        {
+            var analyzers = new DiagnosticAnalyzer[] {new DummyAnalyzer()};
+
+            var testSubject = CreateTestSubject(analyzers: analyzers);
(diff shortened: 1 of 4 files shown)


omnisharp-dotnet/src/SonarLint.OmniSharp.Plugin/DiagnosticWorker/SonarLintDiagnosticWorker.cs
```diff
@@ -56,17 +56,27 @@ internal class SonarLintDiagnosticWorker : CopiedCSharpDiagnosticWorkerWithAnaly
...
             var analysisConfig = sonarLintAnalysisConfigProvider.Get(compilation, workspaceAnalyzerOptions);
             
-            return base.AnalyzeDocument(project, 
+            var result = await base.AnalyzeDocument(project, 
                 analysisConfig.Analyzers, 
                 analysisConfig.Compilation, 
                 analysisConfig.AnalyzerOptions, 
                 document);
+
+            var supportedRules = analysisConfig
```
rita-sonarsource, line 71: I could also take it from `analysisConfig.Compilation.Options.SpecificDiagnosticOptions` but this was easier to setup in the tests. I'm not sure if it makes much difference in terms of performance.

duncanp-sonar: The list of supported rules is fixed, so we don't need to recalculate this value for each analysis.

rita-sonarsource: Hmm, good point. The current design doesn't support this neatly.
So I see 2 options:
1. Take `ISonarAnalyzerCodeActionProvider` out of `SonarLintAnalysisConfigProvider` and back into `SonarLintDiagnosticWorker`: this way the worker could cache the rules in the ctor.
2. Add `SupportedRules` to `AnalysisConfig`, that will use a cached value of analyzers returned by `ISonarAnalyzerCodeActionProvider`. 
3. WDYT?

duncanp-sonar: (2) seems to make more sense - all the analyzer config will be in one place, and the diagnostic worker will only need to know about the single config provider class.

omnisharp-dotnet/src/SonarLint.OmniSharp.Plugin.UnitTests/DiagnosticWorker/SonarLintAnalysisConfigProviderTests.cs
```diff
@@ -146,18 +157,19 @@ private AnalyzerOptions CreateOptions(params AdditionalText[] existingAdditional
 
         private SonarLintAnalysisConfigProvider CreateTestSubject(
             RuleDefinition[] rules = null,
-            ImmutableArray<DiagnosticAnalyzer> analyzers = default,
```
rita-sonarsource, line 149: Minor change: the tests previously passed a default immutable array and this was fine, but now I'm performing linq operations on it to extract the descriptors. This results in an exception `This operation cannot be performed on a default instance of ImmutableArray<T>.  Consider initializing the array, or checking the ImmutableArray<T>.IsDefault property.`

omnisharp-dotnet/src/SonarLint.OmniSharp.Plugin/DiagnosticWorker/SonarLintDiagnosticWorker.cs
rita-sonarsource: Done, please re-review

duncanp-sonar (approved):

PR #67, opened 2021-06-23
Support secondary locations

### LINKED ISSUE #7: Return secondary locations
### Description

Modify the new SonarLint code check to include information about secondary locations.

- [x] modify the data structure returned by the service endpoint
- [x] populate the new structure
- [x] unit tests

rita-sonarsource:
Fixes #7


omnisharp-dotnet/src/SonarLint.OmniSharp.Plugin.UnitTests/DiagnosticWorker/SonarLintDiagnosticWorkerTests.cs
```diff
@@ -158,6 +157,45 @@ public async Task AnalyzeDocument_CompilerWarningsAndErrorsAreIgnored()
             result.Should().BeEmpty();
         }
 
+        [TestMethod]
+        public async Task AnalyzeDocument_IssueContainsAdditionalLocations_ReturnsRoslynAdditionalLocations()
```
rita-sonarsource, line 161: The purpose of this test is to assert that our custom worker doesn't mess with the diagnostics.
I'm not sure if it adds much value and perhaps we should keep it simple and remove it, WDYT?

omnisharp-dotnet/src/SonarLint.OmniSharp.Plugin.UnitTests/SonarLint.OmniSharp.Plugin.UnitTests.csproj
```diff
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <TargetFramework>.netcoreapp3.1</TargetFramework>
-    <LangVersion>8</LangVersion>
+    <LangVersion>9</LangVersion>
```
rita-sonarsource, line 5: For ease of work with Tuple objects in the tests.

omnisharp-dotnet/src/SonarLint.OmniSharp.Plugin/DiagnosticWorker/DiagnosticLocation/CodeLocation.cs
```diff
@@ -0,0 +1,84 @@
...
+        int Column { get; set; }
+        int EndLine { get; set; }
+        int EndColumn { get; set; }
+        string Text { get; set; }
+    }
+
+    /// <summary>
+    /// Based on <see cref="QuickFix"/>
+    /// </summary>
+    internal sealed class CodeLocation : ICodeLocation
```
rita-sonarsource, line 40: This is pretty much a copy-paste of OmniSharp's QuickFix, but without the Projects property.
The `GetHashCode` was auto generated

omnisharp-dotnet/src/SonarLint.OmniSharp.Plugin/DiagnosticWorker/DiagnosticLocation/SonarLintDiagnosticLocation.cs
```diff
@@ -0,0 +1,27 @@
...
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+namespace SonarLint.OmniSharp.Plugin.DiagnosticWorker.DiagnosticLocation
+{
+    internal class SonarLintDiagnosticLocation : global::OmniSharp.Models.Diagnostics.DiagnosticLocation, ICodeLocation
```
rita-sonarsource, line 23: Not sure why there is an assembly alias here. Adding `using OmniSharp.Models.Diagnostics;` doesn't solve it.

omnisharp-dotnet/src/SonarLint.OmniSharp.Plugin/DiagnosticWorker/OmniSharp/Configuration.cs
```diff
@@ -12,6 +12,7 @@ namespace OmniSharp
     /// </summary>
     internal static class Configuration
     {
+        public static bool ZeroBasedIndices = false;
```
rita-sonarsource, line 15: Used in the "new" SonarLintDiagnosticLocation

omnisharp-dotnet/src/SonarLint.OmniSharp.Plugin/DiagnosticWorker/OmniSharp/ZeroBasedIndexConverter.cs
```diff
@@ -0,0 +1,115 @@
...
+using System.Linq;
+using Newtonsoft.Json;
+
+namespace OmniSharp.Models
+{
+    /// <summary>
+    /// Copied unchanged from OmniSharp.Models.ZeroBasedIndexConverter
+    /// https://github.com/OmniSharp/omnisharp-roslyn/blob/master/src/OmniSharp.Abstractions/Models/ZeroBasedIndexConverter.cs
+    /// </summary>
+    internal class ZeroBasedIndexConverter : JsonConverter
```
rita-sonarsource, line 18: Used in the "new" SonarLintDiagnosticLocation

omnisharp-dotnet/src/SonarLint.OmniSharp.Plugin/DiagnosticWorker/OmniSharp/DiagnosticExtensions.cs
```diff
@@ -38,11 +40,12 @@ internal static DiagnosticLocation ToDiagnosticLocation(this Diagnostic diagnost
                     .Descriptor.CustomTags
                     .Where(x => _tagFilter.Contains(x))
                     .ToArray(),
-                Id = diagnostic.Id
+                Id = diagnostic.Id,
+                AdditionalLocations = diagnostic.ToAdditionalLocations()
```
rita-sonarsource, line 44: `ToAdditionalLocations` is implemented in the new SonarLintDiagnosticLocationExtensions

omnisharp-dotnet/src/SonarLint.OmniSharp.Plugin.UnitTests/Services/SonarLintCodeCheckServiceTests.cs
```diff
@@ -30,6 +30,7 @@
 using OmniSharp.Models;
 using OmniSharp.Roslyn.CSharp.Services.Diagnostics;
 using SonarLint.OmniSharp.Plugin.DiagnosticWorker;
+using SonarLint.OmniSharp.Plugin.DiagnosticWorker.DiagnosticLocation;
 using SonarLint.OmniSharp.Plugin.Services;
 
 namespace SonarLint.OmniSharp.Plugin.UnitTests.Services
```
rita-sonarsource, line 36: Currently the flow is as follows:
1. java side invokes `CodeCheckService` --> it invokes our custom diagnostics worker --> worker returns OmniSharp's `DocumentDiagnostics`, that holds an array of Roslyn's `Diagnostic`.
2. `CodeCheckService` then calls the extension method `DiagnosticExtensions.DistinctDiagnosticLocationsByProject`, which calls our new `SonarLintDiagnosticLocationExtensions.ToAdditionalLocations`
3. `SonarLintDiagnosticLocationExtensions` then converts Roslyn's diagnostics to the new `CodeLocation` that has the span + message.

With this structure, it's tough to know what I should be testing in `CodeCheckServiceTests`. 
I modified the tests to test that secondary locations are returned, but I'm also comparing that they're returned correctly (minus the messages) -- which feels wrong, as that's the responsibility of `SonarLintDiagnosticLocationExtensionsTests`.
And it especially feels wrong since I'm only comparing the spans and not the message.

Adding the assertion for messages made the tests overly complicated so I reverted that. 
I think they're still overly complicated, and maybe I should remove the span assertions as well. 
Let me know what you think.

omnisharp-dotnet/src/SonarLint.OmniSharp.Plugin/DiagnosticWorker/DiagnosticLocation/CodeLocation.cs
duncanp-sonar: Auto-generated by what? Worth commenting that it's auto-generated.

omnisharp-dotnet/src/SonarLint.OmniSharp.Plugin/DiagnosticWorker/DiagnosticLocation/CodeLocation.cs
```diff
@@ -0,0 +1,84 @@
...
+        public int Line { get; set; }
+        [JsonConverter(typeof(ZeroBasedIndexConverter))]
+        public int Column { get; set; }
+        [JsonConverter(typeof(ZeroBasedIndexConverter))]
+        public int EndLine { get; set; }
+        [JsonConverter(typeof(ZeroBasedIndexConverter))]
+        public int EndColumn { get; set; }
+        public string Text { get; set; }
+
+        private bool Equals(CodeLocation other)
```
duncanp-sonar, line 53: Are we relying on the equality/hash code implementations being correct?
If so, what level of testing do we need to do for them?

omnisharp-dotnet/src/SonarLint.OmniSharp.Plugin/DiagnosticWorker/DiagnosticLocation/SonarLintDiagnosticLocation.cs
duncanp-sonar: That is a bit of a worry. I'll have a look at it when I do the ticket to change the name of the extension; maybe that will fix it.

omnisharp-dotnet/src/SonarLint.OmniSharp.Plugin.UnitTests/DiagnosticWorker/SonarLintDiagnosticWorkerTests.cs
duncanp-sonar: Hhhmm. It does test a behaviour of the worker, so it has some value. Since you've written it we might as well keep it.
But I'd change the test name - I wouldn't have guessed from the name that the purpose of the test is what you described in your comment above.

omnisharp-dotnet/src/SonarLint.OmniSharp.Plugin.UnitTests/Services/SonarLintCodeCheckServiceTests.cs
duncanp-sonar: I'd keep it simple; the service either returns all diagnostics, or those for a specific file. Testing that should be enough for now.

omnisharp-dotnet/src/SonarLint.OmniSharp.Plugin.UnitTests/DiagnosticWorker/SonarLintDiagnosticWorkerTests.cs
rita-sonarsource: The thing is, we're not testing the other properties of the diagnostics (not even the primary location) -- so it's odd that we're testing just the secondary. I don't know, I feel that the test is out of place.

omnisharp-dotnet/src/SonarLint.OmniSharp.Plugin.UnitTests/Services/SonarLintCodeCheckServiceTests.cs
rita-sonarsource: Ok I will remove the additional location tests. I felt they were too much for CodeCheckServiceTests.

omnisharp-dotnet/src/SonarLint.OmniSharp.Plugin/DiagnosticWorker/DiagnosticLocation/CodeLocation.cs
rita-sonarsource: Hmm, good question. I copied it from OmniSharp, and they rely on it in `DiagnosticExtensions.DistinctDiagnosticLocationsByProject` to filter out duplicate locations.
However, our new `SonarLintDiagnosticLocation` doesn't override the equality functions, so secondary locations won't be part of the distinct check anyway.
Maybe I should then just remove the equality functions -- WDYT?

omnisharp-dotnet/src/SonarLint.OmniSharp.Plugin.UnitTests/Services/SonarLintCodeCheckServiceTests.cs
rita-sonarsource: Edit: I added a helper class, `DiagnosticsToCodeLocationsConverter` and wrote proper tests for it that test the duplication and secondary locations. I think it's cleaner now. Please re-review 🙂

omnisharp-dotnet/src/SonarLint.OmniSharp.Plugin.UnitTests/DiagnosticWorker/SonarLintDiagnosticLocationExtensionsTests.cs
```diff
@@ -0,0 +1,131 @@
...
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.Text;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Moq;
+using SonarLint.OmniSharp.Plugin.DiagnosticWorker.DiagnosticLocation;
+
+namespace SonarLint.OmniSharp.Plugin.UnitTests.DiagnosticWorker
+{
+    [TestClass]
+    public class SonarLintDiagnosticLocationExtensionsTests
```
rita-sonarsource, line 34: I added `DiagnosticsToCodeLocationsConverterTests` that has potential overlap with these tests. Maybe it's better to remove this test suite - WDYT?

omnisharp-dotnet/src/SonarLint.OmniSharp.Plugin/DiagnosticWorker/DiagnosticLocation/CodeLocation.cs
duncanp-sonar: We could make the probably-safe-but-slightly-hacky assumption that if the non-secondary-location parts of the issue match then the secondary-location parts will match too (which is effectively what we are doing at the moment).

You're right, there's no point in implementing the methods here if we don't also implement them in the `SonarLintDiagnosticLocation`. I would say it's ok not to implement it now, and drop these methods.
However, I would add a comment to `SonarLintDiagnosticLocation` saying secondary locations are ignored when comparing.

omnisharp-dotnet/src/SonarLint.OmniSharp.Plugin/Services/DiagnosticsToQuickFixConverter.cs
```diff
@@ -0,0 +1,49 @@
...
+using SonarLint.OmniSharp.Plugin.DiagnosticWorker.DiagnosticLocation;
+
+namespace SonarLint.OmniSharp.Plugin.Services
+{
+    internal interface IDiagnosticsToCodeLocationsConverter
+    {
+        ImmutableArray<SonarLintDiagnosticLocation> Convert(ImmutableArray<DocumentDiagnostics> documentDiagnostics, string fileNameFilter);
+    }
+    
+    [Export(typeof(IDiagnosticsToCodeLocationsConverter)), Shared]
```
duncanp-sonar, line 35: This is effectively a private helper class for the diagnostic worker. Does it need to be exported?

omnisharp-dotnet/src/SonarLint.OmniSharp.Plugin.UnitTests/Services/SonarLintCodeCheckServiceTests.cs
duncanp-sonar: The new helper class has made the tests for the service simpler, but it's made our service harder to compare against the OmniSharp version, and it's introduced the test-duplication problem in `DiagnosticToCodeLocationsConverter` and `SonarLintDiagnosticLocationExtensions`.

I think there are two competing goals: (1) leave the code as close to the original so it's easy to compare, and (2) refactor the code so it's easier to test. We're now somewhere in the middle.

omnisharp-dotnet/src/SonarLint.OmniSharp.Plugin/DiagnosticWorker/OmniSharp/ZeroBasedIndexConverter.cs
duncanp-sonar: Are there tests we could copy too to improve the coverage?

duncanp-sonar (approved):

omnisharp-dotnet/src/SonarLint.OmniSharp.Plugin/DiagnosticWorker/OmniSharp/ZeroBasedIndexConverter.cs
rita-sonarsource: No :(

Issue #33, opened 2021-06-16
Rename .NET assembly to disambiguate it from the Java artefacts


duncanp-sonar:
### Description

New name: `SonarLint.OmniSharp.DotNet.Services`

PR #71, opened 2021-06-24
Rename assembly

duncanp-sonar:
Part of #33

TODO
* changes namespaces and project names

### DIFF
--- a/omnisharp-dotnet/src/SonarLint.OmniSharp.Plugin/SonarLint.OmniSharp.Plugin.csproj
+++ b/omnisharp-dotnet/src/SonarLint.OmniSharp.Plugin/SonarLint.OmniSharp.Plugin.csproj
@@ -5,6 +5,7 @@
 
     <!-- We don't want to ship the OmniSharp assemblies -->
     <CopyOmniSharpToOutput>false</CopyOmniSharpToOutput>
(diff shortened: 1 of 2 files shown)


omnisharp-dotnet/src/SonarLint.OmniSharp.Plugin/SonarLint.OmniSharp.Plugin.csproj
```diff
@@ -5,6 +5,7 @@
 
     <!-- We don't want to ship the OmniSharp assemblies -->
     <CopyOmniSharpToOutput>false</CopyOmniSharpToOutput>
+    <AssemblyName>SonarLint.OmniSharp.DotNet.Services</AssemblyName>
```
duncanp-sonar, line 8: The namespaces etc should be changed too. However, this is the only change that affects the Java side (since they need to know the name of the plugin assembly).

We can make the other changes separately without impacting the Java side.

rita-sonarsource (approved):

omnisharp-dotnet/src/SonarLint.OmniSharp.Plugin/SonarLint.OmniSharp.Plugin.csproj
rita-sonarsource: Arguably we could just leave the other stuff as-is.

Issue #54, opened 2021-06-21
Code cleanup


duncanp-sonar:
### Description

There are a lot of code smells etc in the analysis on Next that should be addressed before shipping.

The solution is now bound so it should be easier to address the issues in the IDE now.

PR #74, opened 2021-06-24
[2] Minor code cleanup

duncanp-sonar:
- low-risk changes e.g. making methods static, changing "new" syntax

Part of #54

### DIFF
--- a/omnisharp-dotnet/src/Services/DiagnosticWorker/SonarAnalyzerCodeActionProvider.cs
+++ b/omnisharp-dotnet/src/Services/DiagnosticWorker/SonarAnalyzerCodeActionProvider.cs
@@ -30,15 +30,10 @@ namespace SonarLint.OmniSharp.DotNet.Services.DiagnosticWorker
     {
     }
 
-    /// <summary>
-    /// Returns Sonar analyzer diagnostics
-    /// </summary>
-    /// <remarks>
-    /// We want this provider to be used only by our custom diagnostic worker and not the "normal" OmniSharp workers,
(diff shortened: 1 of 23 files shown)


omnisharp-dotnet/src/Services/DiagnosticWorker/SonarAnalyzerCodeActionProvider.cs
```diff
@@ -29,7 +29,10 @@ namespace SonarLint.OmniSharp.DotNet.Services.DiagnosticWorker
     internal interface ISonarAnalyzerCodeActionProvider : ICodeActionProvider
     {
     }
-    
+
+    /// <summary>
+    /// Returns Sonar analyzer diagnostics
```
rita-sonarsource, line 34: There is a comment on the interface itself, so I originally didn't place it on the class.
Why `diagnostics` ? It provides `DiagnosticAnalyzer`, not the diagnostics themselves.

rita-sonarsource (approved):

omnisharp-dotnet/src/Services/DiagnosticWorker/SonarAnalyzerCodeActionProvider.cs
duncanp-sonar: An analyzer was complaining about the presence of the `remarks` doc comment without the `summary`, so I added a `summary` just to shut it up.
The other alternative would have been to change it to a non-doc comment quote, I guess.

PR #76, opened 2021-06-24
Merge .NET and Java pipelines

duncanp-sonar:
Part of #2

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -26,7 +26,7 @@ stages:
       - name: buildConfiguration
         value: 'Release'  
       - name: slnPath
-        value: 'omnisharp-dotnet/SonarLint.OmniSharp.Plugin.sln'
+        value: 'omnisharp-dotnet/SonarLint.OmniSharp.DotNet.Services.sln'
 
     jobs:
       - job: dotnet_build_test_test_analyze
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1,21 +1,11 @@
+trigger:
+- main
+- branch-dotnet
+
 pool:
   vmImage: 'ubuntu-latest'
 
-# Temporarily disable the "main" analysis for changes in the dotnet subfolder.
-# Will be removed as part of #2 later.
-pr:
-  branches:
-    exclude:
-    - branch-dotnet
-    - rg/*
-    - dp/*
 
-trigger:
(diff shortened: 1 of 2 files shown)


azure-pipelines.yml
```diff
@@ -30,12 +20,94 @@ resources:
...
+          displayName: 'Download snk file'
+          name: snk
+          inputs:
+            secureFile: SonarSourceSecret.snk
+
+        - powershell: |
+            # Reads the Sonar project version to use from the Java POM file
+
+            # Calculate the POM file path
+            $versionFilePath = "$env:BUILD_SOURCESDIRECTORY\omnisharp-plugin\pom.xml"
```
henryju, line 45: You could use the root pom as well, the version will be the same.

azure-pipelines.yml
```diff
@@ -30,12 +20,94 @@ resources:
...
+            Write-Host "Reading the Sonar project version from '${versionFilePath}' ..."
+
+            # Read the version from the file
+            [xml]$versionProps = Get-Content "$versionFilePath"
+            $sonarProjectVersion = $versionProps.project.parent.version
+            Write-Host "Sonar project version is '${sonarProjectVersion}'"
+
+            # Set the variable to it can be used by other tasks
+            Write-Host "##vso[task.setvariable variable=SONAR_PROJECT_VERSION;]$sonarProjectVersion"
```
henryju, line 54: What will be the use of this `sonarProjectVersion` variable? Because if I understand correctly, you will get a version like `1.0-SNAPSHOT` that is not the "final" version with the build number.

azure-pipelines.yml
duncanp-sonar: Do it matter which? i.e. is one preferable to the other?

azure-pipelines.yml
duncanp-sonar: It's only used to set the project version for Sonar analysis. The version of the assembly currently doesn't changed.

Ideally, the Sonar project version would be the same as for the Maven project. What value does it use?

henryju: The SQ analysis use the unmodified pom version (so 1.0-SNAPSHOT).

duncanp-sonar:
Superceded by https://github.com/SonarSource/sonarlint-omnisharp/pull/95

PR #79, opened 2021-06-25
Update SonarLint.xml to include only enabled rules

rita-sonarsource:


omnisharp-dotnet/src/Services/Rules/SonarLintXml/RulesToSonarLintConfigurationConverter.cs
```diff
@@ -33,17 +33,19 @@ internal class RulesToSonarLintConfigurationConverter : IRulesToSonarLintConfigu
...
-                Parameters = rule.Parameters?.Select(param =>
-                        new SonarLintKeyValuePair
-                        {
-                            Key = param.Key,
-                            Value = param.Value
-                        })
-                    .ToList()
-            }).ToList();
+            var sonarLintRules = rules
+                .Where(x => x.IsEnabled)
```
duncanp-sonar, line 37: I think we should only write out rules that have parameters, regardless of their enabled status. Let me know if you agree, and I'll make the change.

rita-sonarsource: I agree 👍 Yes, please make the change 🙂

rita-sonarsource: @duncanp-sonar I'm closing this PR, as the "IsEnabled" flag is no longer relevant due to #80

PR #80, opened 2021-06-28
Change RuleDefinitionsRepository to contain only active rules

### LINKED ISSUE #84: Modify rules configuration service to accept a list of active rules
### Description

Modify the contract for the rules configuration service so the caller only has to supply the list of rules to be executed and their parameters. Any rules that are not listed should be disabled.

#### Implementation notes
There are a number of rules that should not be run in the IDE and should not appear in the rule configuration UI i.e. security hotspot rules. However, these rules are enabled by default, so it is necessary to explicitly disable them before running the analysis.

This change simplifies things on the Java/UI side. The Java/UI side already does not display security hotspot rules in the UI.
The .NET extension side already knows all of the rules that are exposed by the analyzers.
Simply disabling any rules that are not explicitly enabled by the Java/UI side will have the effect of disabling any hotspot rules.

rita-sonarsource:
Fixes #84

### DIFF
--- a/omnisharp-dotnet/src/Services.UnitTests/Rules/RulesToReportDiagnosticsConverterTests.cs
+++ b/omnisharp-dotnet/src/Services.UnitTests/Rules/RulesToReportDiagnosticsConverterTests.cs
@@ -58,22 +58,21 @@ namespace SonarLint.OmniSharp.DotNet.Services.UnitTests.Rules
         }
 
         [TestMethod]
-        public void Convert_UnrecognizedActiveRule_ActiveRuleIsIgnored()
+        public void Convert_UnrecognizedActiveRule_ArgumentException()
         {
             var existingAnalyzerRules = new[] {"rule1", "rule2"}.ToImmutableHashSet();
             var activeRules = new[]
             {
-                "some unknown rule",
+                "unknown1",
+                "unknown2",
                 "rule1"
             }.ToImmutableHashSet();
 
             var testSubject = CreateTestSubject();
-            var result = testSubject.Convert(activeRules, existingAnalyzerRules);
+            Action act = () => testSubject.Convert(activeRules, existingAnalyzerRules);
 
-            result.Should().NotBeEmpty();
-            result.Count.Should().Be(2);
-            result.Should().NotContainKey("some unknown rule");
-            result.Should().ContainKeys(existingAnalyzerRules);
+            act.Should().Throw<ArgumentException>().And.ParamName.Should().Be("activeRules");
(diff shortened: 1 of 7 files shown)


omnisharp-dotnet/src/Services/Rules/RulesToReportDiagnosticsConverter.cs
```diff
@@ -18,27 +18,37 @@
...
-                        ? ReportDiagnostic.Warn
-                        : ReportDiagnostic.Suppress);
+            if (allRules.IsEmpty)
+            {
+                throw new ArgumentException("No analyzer rules", nameof(allRules));
+            }
+
+            var diagnosticOptions = allRules
+                .ToDictionary(ruleId => ruleId,
+                    ruleId => activeRules.Contains(ruleId)
```
rita-sonarsource, line 49: As we discussed offline, "IsEnabled" flag is removed and instead non-enabled rules are all analyzer rules that are not in the active rules list.

omnisharp-dotnet/src/Services.UnitTests/Rules/RulesToReportDiagnosticsConverterTests.cs
```diff
@@ -30,44 +30,76 @@ namespace SonarLint.OmniSharp.DotNet.Services.UnitTests.Rules
...
+            result.Count.Should().Be(2);
 
-            result.Should().BeEmpty();
+            result["rule1"].Should().Be(RulesToReportDiagnosticsConverter.DisabledRuleSeverity);
+            result["rule2"].Should().Be(RulesToReportDiagnosticsConverter.DisabledRuleSeverity);
         }
 
         [TestMethod]
-        public void Convert_EnabledAndDisabledRules_DiagnosticSeverityIsWarningAndSuppressed()
+        public void Convert_UnrecognizedActiveRule_ActiveRuleIsIgnored()
```
duncanp-sonar, line 61: I think it's worth testing this.

However, it shouldn't happen in practice, and would indicate there's something wrong with our build/integration. It would be pretty difficult to detect, so it might be worth asserting the product code on the .NET side that all of the supplied IDs are recognised.

(a stronger check would be for us to throw, and have an integration test that enables all rules on the Java side).

duncanp-sonar (approved):

omnisharp-dotnet/src/Services.UnitTests/Rules/RulesToReportDiagnosticsConverterTests.cs
rita-sonarsource: Changed to throwing an Argument Exception with unrecognized rule ids.

PR #87, opened 2021-06-29
Filesystem events

henryju:


omnisharp-plugin/src/main/java/org/sonarsource/sonarlint/omnisharp/OmnisharpFileListener.java
```diff
@@ -0,0 +1,90 @@
...
+        } else {
+          omnisharpProtocol.fileChanged(file, OmnisharpProtocol.FileChangeType.CREATE);
+        }
+        break;
+      case DELETED:
+        omnisharpProtocol.fileChanged(file, OmnisharpProtocol.FileChangeType.DELETE);
+        break;
+      case MODIFIED:
+        if (file.getName().endsWith(".sln")) {
+          // Stop the server so that it is restarted during the next analysis and take into account changes to the solution
```
damien-urruty-sonarsource, line 78: Not needed for csproj ?

omnisharp-plugin/src/test/java/org/sonarsource/sonarlint/omnisharp/CSharpPluginTests.java
```diff
@@ -39,7 +39,7 @@ void getExtensions() {
 
     List<?> extensions = context.getExtensions();
 
-    assertThat(extensions).hasSize(9);
+    assertThat(extensions).hasSize(10);
```
damien-urruty-sonarsource, line 42: I know we have this in every plugin, and I wonder if it really is an interesting thing to test

damien-urruty-sonarsource (approved):
:+1:

omnisharp-plugin/src/test/java/org/sonarsource/sonarlint/omnisharp/CSharpPluginTests.java
henryju: It is useful when you have complex logic to conditionally include extensions depending on the context. So here it has a low value, but this is helping me you increase coverage :)

omnisharp-plugin/src/main/java/org/sonarsource/sonarlint/omnisharp/OmnisharpFileListener.java
henryju: @duncanp-sonar tested that only when a new .csproj file is created we should restart. When the .csproj file is changed, OmniSharp will automatically update.
For the .sln, I decided to restart for any changes, just in case. This should not be a frequent case anyway.

duncanp-sonar: The .csproj will be refreshed as long as we send OmniSharp a `filesChanged (changed)` message.

PR #88, opened 2021-06-29
Secondary locations

henryju:


omnisharp-plugin/src/main/java/org/sonarsource/sonarlint/omnisharp/OmnisharpSensor.java
```diff
@@ -156,16 +158,32 @@ private void scanFile(SensorContext context, InputFile f) {
...
+        if (Files.isSameFile(f.file().toPath(), Paths.get(diag.filename))) {
+          NewIssue newIssue = context.newIssue();
+          newIssue
+            .forRule(ruleKey)
+            .at(createLocation(newIssue, diag, f));
+          for (OmnisharpDiagnosticLocation additionalLocation : diag.additionalLocations) {
+            // Only report secondary locations on the same file
+            if (Files.isSameFile(f.file().toPath(), Paths.get(additionalLocation.filename))) {
+              newIssue.addLocation(createLocation(newIssue, additionalLocation, f));
+            }
```
damien-urruty-sonarsource, line 172: I suppose you added this limitation because you can't retrieve an InputFile from the relative path. My concern is that we could get a partial flow, e.g. the numbering in the IDE could start at 1 but not represent the real origin, or have holes in the middle. Can't we declare "incomplete" locations ?

We could probably improve the ModuleFileSystem (or regular FileSystem) to propose a way to retrieve an InputFile based on a relative path

omnisharp-plugin/src/main/java/org/sonarsource/sonarlint/omnisharp/OmnisharpDiagnosticLocation.java
```diff
@@ -0,0 +1,34 @@
...
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonarsource.sonarlint.omnisharp;
+
+import javax.annotation.Nullable;
+
+public class OmnisharpDiagnosticLocation {
```
damien-urruty-sonarsource, line 24: I was going to comment that the "Omnisharp" prefix is probably useless here and realized we have it on every class. I know every plugin is made like that, I think it's redundant

damien-urruty-sonarsource (approved):

omnisharp-plugin/src/main/java/org/sonarsource/sonarlint/omnisharp/OmnisharpSensor.java
henryju: And also because we are not aware of rules having cross file issue locations, so this will be an improvement for later

omnisharp-plugin/src/main/java/org/sonarsource/sonarlint/omnisharp/OmnisharpDiagnosticLocation.java
henryju: I think here it is not redundant. The package "omnisharp" is there because we are in the sonarlint-omnisharp plugin. But this object represents a diagnostic coming from the OmniSharp *server*, so I prefer to be explicit. A matter of taste I think.

henryju: For other classes, you're probably right

Issue #82, opened 2021-06-28
Rules parameters are ignored


duncanp-sonar:
### Description

Rules parameters are passed to the analysers but are not being used. Also, the log is full of warnings about missing files.

#### Background

The underlying problem is a bug in the analyzers - they are trying to read the SonarLint.xml from disc, rather than using the Roslyn API.

There a couple of obvious solutions:
1) change the analyzers to use the Roslyn API.
2) work round the issue in our plugin by saving the file to disc.

Option (2) is superficially attractive, but has a number of problems: 
* creating temporary files has previously caused problems on different OSes e.g. permissions, file locking, virus checking
* analyses can run in parallel, and the configuration could change between each run. We'd either have to create a separate file for each run, or track whether the configuration had changed and write/update a single file and deal with locking/concurrency issues.

rita-sonarsource:
Fixed on the analyzer side: https://github.com/SonarSource/sonar-dotnet/pull/4616

PR #89, opened 2021-06-30
Serialize SonarLint.xml without BOM

rita-sonarsource:
Part of #82

### DIFF
--- a/omnisharp-dotnet/src/Services/Rules/SonarLintXml/SonarLintConfigurationSerializer.cs
+++ b/omnisharp-dotnet/src/Services/Rules/SonarLintXml/SonarLintConfigurationSerializer.cs
@@ -42,7 +42,7 @@ namespace SonarLint.OmniSharp.DotNet.Services.Rules.SonarLintXml
                 Indent = true,
                 NamespaceHandling = NamespaceHandling.OmitDuplicates,
                 OmitXmlDeclaration = false,
-                Encoding = Encoding.UTF8
+                Encoding = new UTF8Encoding(false)
             };
 
             using (var stream = new MemoryStream {Position = 0})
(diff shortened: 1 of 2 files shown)


omnisharp-dotnet/src/Services/Rules/SonarLintXml/SonarLintConfigurationSerializer.cs
```diff
@@ -42,7 +42,7 @@ public string Serialize(SonarLintConfiguration sonarLintConfiguration)
                 Indent = true,
                 NamespaceHandling = NamespaceHandling.OmitDuplicates,
                 OmitXmlDeclaration = false,
-                Encoding = Encoding.UTF8
+                Encoding = new UTF8Encoding(false)
```
rita-sonarsource, line 45: I'm torn how to fix this.
In the [analyzer](https://github.com/SonarSource/sonar-dotnet/blob/be97e12483568f55d25883248f64102e64d5831f/analyzers/src/SonarAnalyzer.Common/Helpers/ParameterLoader.cs#L98), I'm using `XDocument.Parse`, which fails with `XmlException: Data at the root level is invalid` if the file contains BOM. 
This serialization was for some reason generating [ZWNBSP BOM](https://en.wikipedia.org/wiki/Word_joiner) and thus failing on `XDocument.Parse`.

I'm not sure if the fix should be on the analyzer's side.

All the committed SonarLint.xml files on the analyzer's side don't contain BOM, and the generated SonarLint.xml in SLVS also works (in unit tests). But what about the other places that generate SonarLint.xml? I don't want to add unnecessary code to the analyzer but I don't know if there is a problem here.

duncanp-sonar: Does XDocument.Load fail if there is a BOM in the file? If so, then we're not making things any worse with this change.

We don't have any other code that generates SonarLint.xml files that I know of - just the scanner and SLVS.

Fixing it on our side isn't ideal, but you've added a test that checks the file can be reloaded. What's missing is some documentation:
* in the product code, a comment that we're deliberately saving without a BOM, and
* in the test, a comment that the test is using the same method as the analyzer, with a ref to the analyzer code.

duncanp-sonar (approved):

PR #95, opened 2021-07-06
Pipeline: added .NET project build/test/analyse to main pipeline

duncanp-sonar:


duncanp-sonar:
Same changes as in the previous PR #76, except that the Sonar project analysis version is now taken from the root-level `pom.xml`.

henryju (approved):
LGTM!

PR #96, opened 2021-07-06
Rework server/endpoint collaboration

henryju:
The key change is that all synchronized methods are in the same class `OmniSharpServer`, to garantee that no concurrent request can be made while the server is starting/stopping

### DIFF
--- a/omnisharp-plugin/src/main/java/org/sonarsource/sonarlint/omnisharp/OmnisharpServer.java
+++ b/omnisharp-plugin/src/main/java/org/sonarsource/sonarlint/omnisharp/OmnisharpServer.java
@@ -153,6 +153,7 @@ public class OmnisharpServer implements Startable {
 
   private synchronized void checkAlive() {
     if (omnisharpStarted && startedProcess != null && !startedProcess.isAlive()) {
+      LOG.error("Process terminated unexpectedly");
       cleanResources();
     }
   }
@@ -174,12 +175,7 @@ public class OmnisharpServer implements Startable {
     streamConsumer.consumeStream(startedProcess.getErrorStream(), LOG::error);
     output = startedProcess.getOutputStream();
 
-    Instant start = clock.instant();
-    do {
-      if (startLatch.await(1, TimeUnit.SECONDS)) {
-        break;
-      }
-    } while (startedProcess.isAlive() && Duration.between(start, clock.instant()).compareTo(serverStartupMaxWait) < 0);
+    waitForLatchOrProcessDied(startLatch);
     if (!startedProcess.isAlive()) {
       LOG.error("Process terminated unexpectedly");
       cleanResources();
@@ -193,6 +189,8 @@ public class OmnisharpServer implements Startable {
     LOG.info("OmniSharp successfully started");
 
     LOG.info("Waiting for solution/project configuration to be loaded...");
+    waitForLatchOrProcessDied(firstUpdateProjectLatch);
+
     if (!firstUpdateProjectLatch.await(1, TimeUnit.MINUTES)) {
       startedProcess.destroyForcibly();
       throw new IllegalStateException("Timeout waiting for solution/project configuration to be loaded");
@@ -200,6 +198,18 @@ public class OmnisharpServer implements Startable {
     LOG.info("Solution/project configuration loaded");
 
     omnisharpStarted = true;
+  }
+
+  /**
+   * Wait for the latch to be counted down, or the process to died, which one come first.
+   */
+  private void waitForLatchOrProcessDied(CountDownLatch latch) throws InterruptedException {
+    Instant start = clock.instant();
+    do {
+      if (latch.await(1, TimeUnit.SECONDS)) {
+        break;
+      }
+    } while (startedProcess.isAlive() && Duration.between(start, clock.instant()).compareTo(serverStartupMaxWait) < 0);
   }
 
   private String buildPathEnv(@Nullable Path dotnetCliPath) {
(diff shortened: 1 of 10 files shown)


omnisharp-plugin/src/main/java/org/sonarsource/sonarlint/omnisharp/StreamConsumer.java
```diff
@@ -0,0 +1,65 @@
...
+        errorReader.lines().forEach(consumer);
+      } catch (IOException e) {
+        LOG.error("Error while reading stream", e);
+      }
+    });
+  }
+
+  void await() throws InterruptedException {
+    executorService.shutdown();
+    if (!executorService.awaitTermination(2, TimeUnit.SECONDS)) {
```
damien-urruty-sonarsource, line 60: Do we really want to wait up to 2 minutes here ? Would it block something while waiting (i guess the module would be kept alive during this time) ?

omnisharp-plugin/src/main/java/org/sonarsource/sonarlint/omnisharp/OmnisharpServer.java
```diff
@@ -280,51 +277,62 @@ public void start() {
...
       try {
-        closeable.close();
-      } catch (IOException e) {
-        LOG.error("Unable to close", e);
+        if (!startedProcess.waitFor(1, TimeUnit.MINUTES)) {
+          LOG.warn("Unable to terminate OmniSharp, killing it");
+          startedProcess.destroyForcibly();
+        }
+      } catch (InterruptedException e) {
+        Thread.currentThread().interrupt();
```
damien-urruty-sonarsource, line 316: Would this kill the thread that is trying to stop the module originally ? The same call is also in other places

omnisharp-plugin/src/main/java/org/sonarsource/sonarlint/omnisharp/OmnisharpServer.java
```diff
@@ -223,31 +230,23 @@ private String getPathEnv() {
...
-      LOG.debug("Unable to execute command", e);
-      if (result != null) {
-        LOG.debug(result.outputString());
+      int result = CommandExecutor.create().execute(command, output::add, LOG::error, 10_000);
+      if (result != 0) {
+        LOG.debug("Command returned with error: " + command);
+        output.forEach(LOG::debug);
+        return null;
       }
+      return output.isEmpty() ? null : output.get(0);
```
damien-urruty-sonarsource, line 242: Couldn't we lose some values here (all the lines but the first) ?

omnisharp-plugin/src/main/java/org/sonarsource/sonarlint/omnisharp/OmnisharpServer.java
```diff
@@ -136,63 +148,58 @@ public void lazyStart(Path projectBaseDir, @Nullable Path dotnetCliPath) throws
...
+    streamConsumer.consumeStream(startedProcess.getInputStream(), l -> omnisharpResponseProcessor.handleOmnisharpOutput(startLatch, firstUpdateProjectLatch, l));
+    streamConsumer.consumeStream(startedProcess.getErrorStream(), LOG::error);
+    output = startedProcess.getOutputStream();
+
+    Instant start = clock.instant();
+    do {
+      if (startLatch.await(1, TimeUnit.SECONDS)) {
+        break;
+      }
+    } while (startedProcess.isAlive() && Duration.between(start, clock.instant()).compareTo(serverStartupMaxWait) < 0);
```
damien-urruty-sonarsource, line 182: I feel that we have 2 mechanisms for the same thing here, the latch and the "retry" of the latch. Is it necessary ? Couldn't we have an extended latch wait time ? I probably miss something here.

damien-urruty-sonarsource: Also I find the second condition of the while a bit hard to read/not obvious

damien-urruty-sonarsource (approved):
LGTM, a few questions/remarks but no blocker

omnisharp-plugin/src/main/java/org/sonarsource/sonarlint/omnisharp/StreamConsumer.java
henryju: 2 *seconds* :)

omnisharp-plugin/src/main/java/org/sonarsource/sonarlint/omnisharp/OmnisharpServer.java
henryju: No, it will not kill the thread. It will simply set the interrupt flag back, so that it can be checked later.

omnisharp-plugin/src/main/java/org/sonarsource/sonarlint/omnisharp/OmnisharpServer.java
henryju: This is the goal of this simple command execution, return only the first line.

omnisharp-plugin/src/main/java/org/sonarsource/sonarlint/omnisharp/OmnisharpServer.java
henryju: The problem here is that I want to wait for 2 things, and the one that come first:
* when the startLatch is triggered
* or when the process died
Previously I was using a longer await on the startLatch (1 minute), but then it means if the process dies immediately (wrong command line for example), you have to wait for 1 minute before realizing.

PR #99, opened 2021-11-30
Ignore unrecognized active rules

duncanp-sonar:
Fixes SLI-637 (https://jira.sonarsource.com/browse/SLI-637)

### DIFF
--- a/omnisharp-dotnet/src/Services/DiagnosticWorker/SonarLintAnalysisConfigProvider.cs
+++ b/omnisharp-dotnet/src/Services/DiagnosticWorker/SonarLintAnalysisConfigProvider.cs
@@ -58,18 +58,19 @@ namespace SonarLint.OmniSharp.DotNet.Services.DiagnosticWorker
 
         [ImportingConstructor]
         public SonarLintAnalysisConfigProvider(IActiveRuleDefinitionsRepository activeRuleDefinitionsRepository,
-            ISonarAnalyzerCodeActionProvider sonarAnalyzerCodeActionProvider)
+            ISonarAnalyzerCodeActionProvider sonarAnalyzerCodeActionProvider,
+            IRulesToReportDiagnosticsConverter rulesToReportDiagnosticsConverter)
             : this(activeRuleDefinitionsRepository,
                 sonarAnalyzerCodeActionProvider,
-                new RulesToAdditionalTextConverter(),
-                new RulesToReportDiagnosticsConverter())
+                rulesToReportDiagnosticsConverter,
+                new RulesToAdditionalTextConverter())
         {
         }
 
         internal SonarLintAnalysisConfigProvider(IActiveRuleDefinitionsRepository activeRuleDefinitionsRepository,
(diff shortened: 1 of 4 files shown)


omnisharp-dotnet/src/Services/Rules/RulesToReportDiagnosticsConverter.cs
```diff
@@ -48,9 +63,9 @@ internal class RulesToReportDiagnosticsConverter : IRulesToReportDiagnosticsConv
             
             if (unrecognizedActiveRules.Any())
             {
-                throw new ArgumentException($@"Unrecognized active rules: {string.Join(",", unrecognizedActiveRules)}", nameof(activeRules));
+                _logger.LogInformation($@"Unrecognized active rules: {string.Join(", ", unrecognizedActiveRules)}. These might be new rules that exist on the server but not in SonarLint.");
```
duncanp-sonar, line 66: The output looks like this:

image

omnisharp-dotnet/src/Services/DiagnosticWorker/SonarLintAnalysisConfigProvider.cs
```diff
@@ -58,18 +58,19 @@ internal class SonarLintAnalysisConfigProvider : ISonarLintAnalysisConfigProvide
...
         [ImportingConstructor]
         public SonarLintAnalysisConfigProvider(IActiveRuleDefinitionsRepository activeRuleDefinitionsRepository,
-            ISonarAnalyzerCodeActionProvider sonarAnalyzerCodeActionProvider)
+            ISonarAnalyzerCodeActionProvider sonarAnalyzerCodeActionProvider,
+            IRulesToReportDiagnosticsConverter rulesToReportDiagnosticsConverter)
             : this(activeRuleDefinitionsRepository,
                 sonarAnalyzerCodeActionProvider,
-                new RulesToAdditionalTextConverter(),
-                new RulesToReportDiagnosticsConverter())
+                rulesToReportDiagnosticsConverter,
```
duncanp-sonar, line 65: The diagnostics converter is now MEF-imported, so it should appear in the argument list before the non-imported additional text handler.

omnisharp-dotnet/src/Services/Rules/RulesToReportDiagnosticsConverter.cs
```diff
@@ -48,9 +63,9 @@ internal class RulesToReportDiagnosticsConverter : IRulesToReportDiagnosticsConv
             
             if (unrecognizedActiveRules.Any())
             {
-                throw new ArgumentException($@"Unrecognized active rules: {string.Join(",", unrecognizedActiveRules)}", nameof(activeRules));
```
duncanp-sonar, line 51: Fixing the bug is simple - just deleting lines 49-52 would have done.

However, it's not clear why some rules are not being run, so to help customers (and us) work out why the results are different the unknown rule IDs are logged the SonarLint output pane.

PR #103, opened 2022-01-06
Allow to pass mono path

henryju:


omnisharp-plugin/src/main/java/org/sonarsource/sonarlint/omnisharp/OmnisharpServer.java
```diff
@@ -134,17 +135,17 @@ public OmnisharpServer(System2 system2, OmnisharpServicesExtractor servicesExtra
     this.sonarLintRuntime = sonarLintRuntime;
   }
 
-  public synchronized void lazyStart(Path projectBaseDir, @Nullable Path dotnetCliPath) throws IOException, InterruptedException {
+  public synchronized void lazyStart(Path projectBaseDir, @Nullable Path dotnetCliPath, @Nullable Path monoPath) throws IOException, InterruptedException {
     checkAlive();
     if (omnisharpStarted) {
-      if (!Objects.equals(cachedProjectBaseDir, projectBaseDir) || !Objects.equals(cachedDotnetCliPath, dotnetCliPath)) {
-        LOG.info("Using a different project basedir or dotnet CLI path, OmniSharp has to be restarted");
+      if (!Objects.equals(cachedProjectBaseDir, projectBaseDir) || !Objects.equals(cachedDotnetCliPath, dotnetCliPath) || !Objects.equals(cachedMonoPath, monoPath)) {
```
rita-sonarsource, line 141: where is cachedMonoPath being updated? doesn't this always fail?

omnisharp-plugin/src/main/java/org/sonarsource/sonarlint/omnisharp/OmnisharpServer.java
```diff
@@ -256,11 +257,14 @@ static String runSimpleCommand(Command command) {
...
 
-  private ProcessBuilder buildOmnisharpCommand(Path projectBaseDir, String omnisharpLoc) {
+  private ProcessBuilder buildOmnisharpCommand(Path projectBaseDir, String omnisharpLoc, @Nullable Path monoPath) {
     Path omnisharpPath = Paths.get(omnisharpLoc);
     List<String> args = new ArrayList<>();
     if (system2.isOsWindows()) {
       args.add(omnisharpPath.resolve(omnisharpExeWin).toString());
+    } else if (monoPath != null) {
+      args.add(monoPath.toString());
+      args.add("omnisharp/OmniSharp.exe");
```
rita-sonarsource, line 267: why the `omnisharp/OmniSharp.exe` ? I'm not sure what the command is supposed to look like

omnisharp-plugin/src/test/java/org/sonarsource/sonarlint/omnisharp/OmnisharpServerTests.java
```diff
@@ -262,10 +266,46 @@ public Void answer(InvocationOnMock invocation) throws Throwable {
...
+      }
+    }).when(endpoints).stopServer();
+
+    underTest.lazyStart(solutionDir, null, monoPath);
+    verify(endpoints).stopServer();
+    if (SystemUtils.IS_OS_WINDOWS) {
+      // On Windows we don't use the Mono path to run Omnisharp
+      verify(responseProcessor, times(2)).handleOmnisharpOutput(any(), any(), eq("Foo"));
+    } else {
+      verify(responseProcessor, times(1)).handleOmnisharpOutput(any(), any(), eq("Foo"));
```
rita-sonarsource, line 305: Does this check that the server is callable after the restart? I believe there is a bug since cachedMonoPath is never updated so the server will restart forever

rita-sonarsource (commented):
@henryju left comments regarding cachedMonoPath

omnisharp-plugin/src/main/java/org/sonarsource/sonarlint/omnisharp/OmnisharpServer.java
henryju: This is the way to run OmniSharp with a custom Mono. See https://github.com/OmniSharp/omnisharp-vscode/blob/866918a4e0386f2e99e23bc16c2512f153c29062/src/omnisharp/OmnisharpManager.ts#L89

omnisharp-plugin/src/main/java/org/sonarsource/sonarlint/omnisharp/OmnisharpServer.java
henryju: Good catch!

omnisharp-plugin/src/test/java/org/sonarsource/sonarlint/omnisharp/OmnisharpServerTests.java
henryju: I will update the test

rita-sonarsource (approved):

PR #115, opened 2022-04-05
Hardening

henryju:


omnisharp-plugin/src/main/java/org/sonarsource/sonarlint/omnisharp/OmnisharpServer.java
```diff
@@ -275,6 +294,11 @@ private ProcessBuilder buildOmnisharpCommand(Path projectBaseDir, String omnisha
       args.add("--hostPID");
       args.add(Long.toString(sonarLintRuntime.getClientPid()));
     }
+    if (msBuildPath != null) {
+      args.add("MsBuild:MSBuildOverride:MSBuildPath=" + msBuildPath.toString());
+    } else {
+      args.add("MsBuild:useBundledOnly=true");
+    }
```
damien-urruty-sonarsource, line 301: Before we had nothing so is this parameter required ?

its/src/test/projects/DotNet6Project/DotNet6Project/Program.cs
```diff
@@ -0,0 +1,4 @@
+﻿// See https://aka.ms/new-console-template for more information
+
+// TODO foo
+Console.WriteLine("Hello, World!");
```
damien-urruty-sonarsource, line 4: Is this file needed ? In the test we declare the same content for the input file

damien-urruty-sonarsource (approved):

omnisharp-plugin/src/main/java/org/sonarsource/sonarlint/omnisharp/OmnisharpServer.java
henryju: Yes, it is! :) This is the main point of this change. The ability to pass a custom path will only be used for advanced use cases.

PR #119, opened 2022-06-24
Upgrade to OmniSharp 1.39.0

rita-sonarsource:


omnisharp-dotnet/src/Services/DiagnosticWorker/OmniSharp/CopiedCSharpDiagnosticWorkerWithAnalyzers.cs
```diff
@@ -29,48 +28,58 @@ namespace OmniSharp.Roslyn.CSharp.Services.Diagnostics
...
         private readonly ImmutableArray<ICodeActionProvider> _providers;
         private readonly DiagnosticEventForwarder _forwarder;
         private readonly OmniSharpOptions _options;
         private readonly OmniSharpWorkspace _workspace;
-
+        
         // This is workaround.
         // Currently roslyn doesn't expose official way to use IDE analyzers during analysis.
         // This options gives certain IDE analysis access for services that are not yet publicly available.
         private readonly ConstructorInfo _workspaceAnalyzerOptionsConstructor;
```
rita-sonarsource, line 53: In their latest version they've removed this hacky workaround, but they're using assemblies that are internal and we cannot access the types here. So we are stuck with the hacky old workaround.

omnisharp-dotnet/src/Services/DiagnosticWorker/OmniSharp/CopiedCSharpDiagnosticWorkerWithAnalyzers.cs
```diff
@@ -29,48 +28,58 @@ namespace OmniSharp.Roslyn.CSharp.Services.Diagnostics
...
-                .GetConstructor(new Type[] { typeof(AnalyzerOptions), typeof(Solution) })
-                ?? throw new InvalidOperationException("Could not resolve 'Microsoft.CodeAnalysis.Diagnostics.WorkspaceAnalyzerOptions' for IDE analyzers.");
-
+            // SONAR LINT HACK START
+            var msFeaturesAssembly = Assembly.Load("Microsoft.CodeAnalysis.Features");
+            var ideAnalyzerOptionsType = msFeaturesAssembly.GetType("Microsoft.CodeAnalysis.Diagnostics.IdeAnalyzerOptions");
+            
+            _workspaceAnalyzerOptionsConstructor = msFeaturesAssembly
+                                                       .GetType("Microsoft.CodeAnalysis.Diagnostics.WorkspaceAnalyzerOptions")
+                                                       .GetConstructor(new Type[] { typeof(AnalyzerOptions), typeof(Solution) , ideAnalyzerOptionsType })
```
rita-sonarsource, line 79: Unfortunately the ctor also changed as part of their refactoring, and is now requiring a third argument of IdeAnalyzerOptions, which is also internal and cannot be created. 

From decompiling, it seems to be related to csharp built-in code refactorings, which we don't use anyway, 

image
image
image

omnisharp-dotnet/src/Services/DiagnosticWorker/OmniSharp/CopiedCSharpDiagnosticWorkerWithAnalyzers.cs
```diff
@@ -201,66 +248,87 @@ private void OnWorkspaceChanged(object sender, WorkspaceChangeEventArgs changeEv
                 case WorkspaceChangeKind.SolutionReloaded:
                     QueueDocumentsForDiagnostics();
                     break;
-
             }
         }
 
+        private AnalyzerOptions CreateAnalyzerOptions(Project project)
+        {
+            var workspaceAnalyzerOptions = (AnalyzerOptions)_workspaceAnalyzerOptionsConstructor.Invoke(new object[] { project.AnalyzerOptions, project.Solution, null });
```
rita-sonarsource, line 256: This is different from the previous hack: I'm now passing the third mandatory argument, but I'm always passing null because it seems to be related to code refactorings and we don't use it. Tests pass without it. 
See comments above for the screenshots what their code is actually doing.

omnisharp-dotnet/src/Services/SonarLint.OmniSharp.DotNet.Services.csproj
```diff
@@ -5,19 +5,27 @@
 
 		<!-- We don't want to ship the OmniSharp assemblies -->
 		<CopyOmniSharpToOutput>false</CopyOmniSharpToOutput>
+		<LangVersion>9</LangVersion>
```
rita-sonarsource, line 8: Required by the copied code

omnisharp-dotnet/src/Services/SonarLint.OmniSharp.DotNet.Services.csproj
```diff
@@ -5,19 +5,27 @@
...
 	</PropertyGroup>
 
 	<ItemGroup>
 		<!-- We need to target the same Roslyn version as is used by OmniSharp.
          Referencing a different version of NewtonSoft.Json doesn't seem to cause a problem,
          but we're targeting the same version to be on the safe side.
     -->
-		<PackageReference Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="4.0.0-6.final" />
+		<PackageReference Include="Microsoft.CodeAnalysis.CSharp.Features" Version="4.3.0-1.final" />
```
rita-sonarsource, line 16: For some reason, the tests failed at runtime until I added all of these dependencies explicitly.

its/src/test/java/org/sonarsource/sonarlint/omnisharp/its/OmnisharpIntegrationTests.java
```diff
@@ -61,6 +62,7 @@
 import static org.assertj.core.api.Assertions.tuple;
 import static org.junit.jupiter.api.Assertions.fail;
 
+@Disabled("Tests need to reference the .net 6 build of OmniSharp")
```
rita-sonarsource, line 65: @henryju will do the fix on the java side. The tests currently fail because we're downloading the .net framework version of omnisharp, which now fails to analyze any project that is NOT old-sdk.

duncanp-sonar (approved):
Looks good. Just a question about whether we really need to create a "real" `WorkspaceAnalyzerOptions` instance for our analyzers to run correctly.

rita-sonarsource:
@duncanp-sonar to unblock @henryju I will merge this PR and take a look at the reflection hack in a separate commit.

PR #125, opened 2022-07-07
Add a property to control loadProjectOnDemand

henryju:


damien-urruty-sonarsource (approved):
Will those new properties we introduced during the iteration be documented somewhere, on the wiki maybe ?
Also it might make the code easier if we implemented [SLCORE-319](https://jira.sonarsource.com/browse/SLCORE-319).

henryju:
Yes, I guess we will document them in the wiki. And definitely +1 for SLCORE-319

PR #136, opened 2022-11-01
Populate quick fixes in the API response

rita-sonarsource:


omnisharp-dotnet/src/Services/DiagnosticWorker/OmniSharp/CopiedCSharpDiagnosticWorker.cs
```diff
@@ -137,10 +141,18 @@ private async Task<DiagnosticResult> ProcessNextItem(string filePath)
...
 
+            var codeLocations = new List<DiagnosticLocation>();
+
+            foreach (var diagnostic in items)
+            {
+                var quickFixes = await _diagnosticQuickFixesProvider.GetDiagnosticQuickFixes(diagnostic, filePath);
+                codeLocations.Add(diagnostic.ToDiagnosticLocation(quickFixes));
+            }
+
             return new DiagnosticResult()
```
rita-sonarsource, line 152: I'm not sure that we need this. Seems that this code flow is executed only when the LSP calls it: https://github.com/OmniSharp/omnisharp-roslyn/blob/ff21ac475cdf8afb15579f2b5fb346945db2a7e9/src/OmniSharp.LanguageServerProtocol/Eventing/LanguageServerEventEmitter.cs#L34

I don't think we're using anything like that on the SL side, and even if we were, the emitter code seems to suggest that it won't be translated anyway: https://github.com/OmniSharp/omnisharp-roslyn/blob/3388b9b5038ecc380091954d985268a4398b5676/src/OmniSharp.LanguageServerProtocol/Helpers.cs#L19

I'm considering dropping this code from here. WDYT?

omnisharp-dotnet/src/Services.UnitTests/Services/DiagnosticsToCodeLocationsConverterTests.cs
```diff
@@ -136,18 +140,72 @@ public void Convert_HasAdditionalLocations_ReturnsWithAdditionalLocations(string
         [TestMethod]
         [DataRow(null)]
         [DataRow("file1.cs")]
-        public void Convert_HasDuplicatedDiagnostics_ReturnsDistinctLocations(string fileName)
+        public async Task Convert_HasQuickFixes_ReturnsWithQuickFixes(string fileName)
```
rita-sonarsource, line 143: New tests

omnisharp-dotnet/src/Services.UnitTests/Services/DiagnosticsToCodeLocationsConverterTests.cs
```diff
@@ -191,15 +250,13 @@ private static Diagnostic CreateDiagnostic(FileLinePositionSpan locationSpan, pa
...
             var additionalLocationsMessages = GetLocationMessages(additionalLocationsWithMessages);
 
-            var diagnostic = new Mock<Diagnostic>();
-            diagnostic.SetupGet(x => x.Id).Returns(locationSpan.GetHashCode().ToString);
-            diagnostic.SetupGet(x => x.Location).Returns(location.Object);
-            diagnostic.SetupGet(x => x.Severity).Returns(DiagnosticSeverity.Info);
-            diagnostic.SetupGet(x => x.Descriptor).Returns(descriptor);
-            diagnostic.SetupGet(x => x.AdditionalLocations).Returns(additionalLocations);
-            diagnostic.SetupGet(x => x.Properties).Returns(additionalLocationsMessages);
+            var diagnostic = Diagnostic.Create(descriptor,
```
rita-sonarsource, line 253: I don't know how it worked before but it doesn't now. You can't mock Diagnostic. I don't understand what this did before but now it returns null when you call diagnostic.Object.

I changed it to use the normal Create method which creates a real diagnostic but with same arguments as before.

omnisharp-dotnet/src/Services.UnitTests/Services/DiagnosticsToCodeLocationsConverterTests.cs
```diff
@@ -177,8 +235,9 @@ private static void AssertLocation(ICodeLocation codeLocation, FileLinePositionS
 
         private static Diagnostic CreateDiagnostic(FileLinePositionSpan locationSpan, params (Location, string)[] additionalLocationsWithMessages)
         {
-            var location = new Mock<Location>();
-            location.Setup(x => x.GetMappedLineSpan()).Returns(locationSpan);
+            var location = Location.Create(locationSpan.Path,
```
rita-sonarsource, line 238: See comment below regarding mocking. I don't understand how this worked. location.Object returns null...

So I replaced it with non-mocks.

omnisharp-dotnet/src/Services.UnitTests/Services/DiagnosticsToCodeLocationsConverterTests.cs
```diff
@@ -136,18 +140,72 @@ public void Convert_HasAdditionalLocations_ReturnsWithAdditionalLocations(string
...
+            var convertedLocation3 = result[2];
+            convertedLocation3.QuickFixes.Should().NotBeNull();
+            convertedLocation3.QuickFixes.Length.Should().Be(1);
+            convertedLocation3.QuickFixes[0].Should().Be(quickFix2);
+        }
+
+        [TestMethod]
+        [DataRow(null)]
+        [DataRow("file1.cs")]
+        public async Task Convert_HasDuplicatedDiagnostics_ReturnsDistinctLocations(string fileName)
```
rita-sonarsource, line 194: This test used to pass but when I changed to non-mocked diagnostic, it started failing.
I'm not sure it was ever correct. Omnisharp's code seems to filter by distinct diagnostics, not distinct locations: https://github.com/OmniSharp/omnisharp-roslyn/blob/ff21ac475cdf8afb15579f2b5fb346945db2a7e9/src/OmniSharp.Roslyn.CSharp/Helpers/DiagnosticExtensions.cs#L44

omnisharp-dotnet/src/Services/DiagnosticWorker/OmniSharp/CopiedCSharpDiagnosticWorker.cs
duncanp-sonar: Our original approach for the first release was to leave this class as untouched as possible to make it easy to diff against  the "original" in OmniSharp i.e. so we could easily see if the "original" has changes.
The only changes we made are to make a couple of methods `virtual` so we can override part of the behaviour.

We've added more features since then. Does our original approach still make sense, or are there now reasons to make changes to in this file, rather than our derived class?

omnisharp-dotnet/src/Services.UnitTests/Services/DiagnosticsToCodeLocationsConverterTests.cs
duncanp-sonar: > now it returns null when you call diagnostic.Object.

I think this a bug in how VS is displaying information about the object in the IDE: if you compare it against `null` it returns `false`.

image

(having said that, I don't think there's any particular advantage in using a mock in this case)

rita-sonarsource: Oh. Interesting.

omnisharp-dotnet/src/Services/DiagnosticWorker/OmniSharp/CopiedCSharpDiagnosticWorker.cs
```diff
@@ -134,13 +136,12 @@ private async Task<DiagnosticResult> ProcessNextItem(string filePath)
...
-            var items = semanticModels
-                .SelectMany(sm => sm.GetDiagnostics());
+            var items = semanticModels.SelectMany(sm => sm.GetDiagnostics());
 
-            return new DiagnosticResult()
+            return new DiagnosticResult
             {
                 FileName = filePath,
-                QuickFixes = items.Select(x => x.ToDiagnosticLocation()).Distinct().ToArray()
+                QuickFixes = items.Select(x => x.ToDiagnosticLocation(Array.Empty<IQuickFix>())).Distinct().ToArray()
```
rita-sonarsource, line 144: As we discussed, passing an empty list just to make the code compile. This branch of code is not expected to be called in the SL execution path.

omnisharp-dotnet/src/Services/DiagnosticWorker/OmniSharp/CopiedCSharpDiagnosticWorker.cs
rita-sonarsource: As we discussed offline, I believe this chunk of code is not needed at all, as it's not part of SL code execution.

omnisharp-dotnet/src/Services/Services/DiagnosticsToQuickFixConverter.cs
```diff
@@ -19,29 +19,39 @@
...
+
+        public async Task<ImmutableArray<SonarLintDiagnosticLocation>> Convert(ImmutableArray<DocumentDiagnostics> documentDiagnostics, string fileNameFilter)
         {
-            var diagnosticLocations = documentDiagnostics
+             var diagnosticLocations = await documentDiagnostics
                 .Where(x => string.IsNullOrEmpty(fileNameFilter) || x.DocumentPath == fileNameFilter)
-                .DistinctDiagnosticLocationsByProject()
-                .Where(x => x.FileName != null)
-                .ToImmutableArray();
+                .DistinctDiagnosticLocationsByProject(quickFixesProvider);
```
duncanp-sonar, line 52: The overall processing order makes sense (fetch the diagnostics, then fetch the quick fixes for each one).

However, the way the processing is split up among the classes feels odd now - I wouldn't expect a  "converter" class to be making calls to a back-end service to perform more processing and fetch more data.

We should have a think about refactoring this at some point, but not as part of this sprint.

duncanp-sonar (approved):

omnisharp-dotnet/src/Services/Services/DiagnosticsToQuickFixConverter.cs
rita-sonarsource: Originally it wasn't there, it was in the controller. However, I thought I needed it to be part of the Copied worker, so I moved it to lower level. Now that it's not part of the worker anymore, it can be moved higher up again.

omnisharp-dotnet/src/Services/DiagnosticWorker/QuickFixes/DiagnosticQuickFixesProvider.cs
```diff
@@ -101,19 +100,19 @@ public async Task<IQuickFix[]> GetDiagnosticQuickFixes(Diagnostic diagnostic, st
                 var solution = workspace.CurrentSolution;
```
ugras-ergun-sonarsource, line 100: I do not think we need this variable anymore it's used only once

omnisharp-dotnet/src/Services/DiagnosticWorker/QuickFixes/DiagnosticQuickFixesProvider.cs
```diff
@@ -101,19 +100,19 @@ public async Task<IQuickFix[]> GetDiagnosticQuickFixes(Diagnostic diagnostic, st
...
-                foreach (var operation in operations.OfType<ApplyChangesOperation>())
+                if (operations.Length > 1 || operation == null)
                 {
-                    var solutionAfterOperation = operation.ChangedSolution;
-                    var fileChangesResult = await GetFileChangesAsync(solutionAfterOperation, solution, directory, true, false);
-
-                    Debug.Assert(fileChangesResult.FileChanges.All(c => c is ModifiedFileResponse));
+                    continue;
+                }
+                var solutionAfterOperation = operation.ChangedSolution;
```
ugras-ergun-sonarsource, line 109: like before I do not think we need this variable anymore it's used only once

ugras-ergun-sonarsource (approved):
We can clean up some stuff after the changes.

omnisharp-dotnet/src/Services/DiagnosticWorker/QuickFixes/DiagnosticQuickFixesProvider.cs
```diff
@@ -126,9 +126,11 @@ public async Task<IQuickFix[]> GetDiagnosticQuickFixes(Diagnostic diagnostic, st
                 var directory = Path.GetDirectoryName(filePath);
                 var operations = await action.GetOperationsAsync(CancellationToken.None);
 
+                Debug.Assert(operations.Length <= 1, "Expecting quick fixes to contain one or zero operations.");
```
rita-sonarsource, line 129: @duncanp-sonar FYI changed to debug assert as we discussed

PR #138, opened 2022-11-02
Add SonarLintDiagnosticCodeActionsProvider

rita-sonarsource:


omnisharp-dotnet/src/Services.UnitTests/DiagnosticWorker/QuickFixes/DiagnosticCodeActionsProviderTests.cs
```diff
@@ -0,0 +1,141 @@
...
+            var document = CreateDocument();
+
+            var result = await testSubject.GetCodeActions(diagnostic, document);
+
+            return result;
+        }
+
+        private Diagnostic CreateDiagnostic() => Diagnostic.Create(Descriptor, null);
+
+        private Document CreateDocument()
```
rita-sonarsource, line 124: This is the reason for extracting the helper methods.
We can't mock roslyn's Document and it's not easy to create it. I extracted a method that we used in other tests, it also creates a workspace which is not needed but it's an easy way to get a Document object.

omnisharp-dotnet/src/Services/DiagnosticWorker/QuickFixes/DiagnosticCodeActionsProvider.cs
```diff
@@ -0,0 +1,72 @@
...
+        [ImportingConstructor]
+        public DiagnosticCodeActionsProvider([ImportMany] IEnumerable<ISonarAnalyzerCodeActionProvider> codeActionProviders)
+        {
+            codeFixProviders = codeActionProviders.SelectMany(x => x.CodeFixProviders).ToArray();
+        }
+
+        /// <summary>
+        /// Based on <see cref="BaseCodeActionService{TRequest,TResponse}.AppendFixesAsync"/>
+        /// </summary>
+        public async Task<List<CodeAction>> GetCodeActions(Diagnostic diagnostic, Document document)
```
rita-sonarsource, line 55: This will only be used in the new QuickFixesProvider, but this code is ugly and will be easier to mock it.

omnisharp-dotnet/src/Services/DiagnosticWorker/QuickFixes/DiagnosticCodeActionsProvider.cs
```diff
@@ -0,0 +1,72 @@
...
+        {
+            var applicableFixProviders = codeFixProviders.Where(x => x.FixableDiagnosticIds.Any(id => id == diagnostic.Id));
+
+            var actions = new List<CodeAction>();
+
+            foreach (var codeFixProvider in applicableFixProviders)
+            {
+                var context = new CodeFixContext(document, diagnostic, (action, _) => actions.Add(action), default);
+                await codeFixProvider.RegisterCodeFixesAsync(context);
```
rita-sonarsource, line 64: The usage of this is very cumbersome: you're supposed to call RegisterCodeFixesAsync and pass it a context object with a delegate, the class will call the delegate for every action, which we can then store and return... I don't think there is a different way to do it. Copy-pasted and adapted from Omnisharp: https://github.com/OmniSharp/omnisharp-roslyn/blob/80d7f26b258853cafacea773276521319f3a5786/src/OmniSharp.Roslyn.CSharp/Services/Refactoring/V2/BaseCodeActionService.cs#L138

omnisharp-dotnet/src/Services/DiagnosticWorker/QuickFixes/DiagnosticCodeActionsProvider.cs
```diff
@@ -0,0 +1,72 @@
...
+            }
+
+            // todo: do we need `action.CodeAction.GetNestedCodeActions` ? https://github.com/OmniSharp/omnisharp-roslyn/blob/80d7f26b258853cafacea773276521319f3a5786/src/OmniSharp.Roslyn.CSharp/Services/Refactoring/V2/BaseCodeActionService.cs#L230
```
rita-sonarsource, line 67: I don't know what is this and if we need it.

omnisharp-dotnet/src/Services.UnitTests/DiagnosticWorker/OmniSharpWorkspaceHelper.cs
```diff
@@ -0,0 +1,93 @@
...
+using OmniSharp.Services;
+using OmniSharp;
+using Microsoft.CodeAnalysis.Text;
+using System.Threading;
+
+namespace SonarLint.OmniSharp.DotNet.Services.UnitTests.DiagnosticWorker
+{
+    internal static class OmniSharpWorkspaceHelper
+    {
+        public static OmniSharpWorkspace CreateOmnisharpWorkspaceWithDocument(string documentFileName, string documentContent)
```
duncanp-sonar, line 38: Pity they don't use interfaces for their types :-(

omnisharp-dotnet/src/Services.UnitTests/DiagnosticWorker/OmniSharpWorkspaceHelper.cs
```diff
@@ -0,0 +1,93 @@
...
+using Moq;
+using OmniSharp.FileWatching;
+using OmniSharp.Services;
+using OmniSharp;
+using Microsoft.CodeAnalysis.Text;
+using System.Threading;
+
+namespace SonarLint.OmniSharp.DotNet.Services.UnitTests.DiagnosticWorker
+{
+    internal static class OmniSharpWorkspaceHelper
```
duncanp-sonar, line 36: It looks like this code just moved from the existing tests - is that correct?

rita-sonarsource: Yes 👍 Moved and unchanged

omnisharp-dotnet/src/Services.UnitTests/DiagnosticWorker/QuickFixes/DiagnosticCodeActionsProviderTests.cs
```diff
@@ -0,0 +1,141 @@
...
+            var result = await GetCodeActions(codeActionProvider.Object);
+
+            result.Should().BeEmpty();
+        }
+
+        [TestMethod]
+        public async Task GetCodeActions_NoApplicableCodeFixProviders_EmptyList()
+        {
+            var codeFixProvider = new Mock<CodeFixProvider>();
+            codeFixProvider.Setup(x => x.FixableDiagnosticIds).Returns(ImmutableArray<string>.Empty);
```
duncanp-sonar, line 66: This is a fairly weak test that the filtering based on diagnostic id excludes non-matching ids; it would be better if it was a non-empty list with a non-matching diagnostic id.

omnisharp-dotnet/src/Services.UnitTests/DiagnosticWorker/QuickFixes/DiagnosticCodeActionsProviderTests.cs
```diff
@@ -0,0 +1,141 @@
...
+            result.Should().BeEmpty();
+        }
+
+        [TestMethod]
+        public async Task GetCodeActions_HasApplicableCodeFixProvider_ActionsRegistered()
+        {
+            var codeFixProvider = new Mock<CodeFixProvider>();
+            codeFixProvider
+                .Setup(x => x.FixableDiagnosticIds)
+                .Returns(new[] {Descriptor.Id}.ToImmutableArray);
```
duncanp-sonar, line 84: This test relies the `GetCodeActions` method using this specific diagnostic id. This assumption is not obvious when reading the test; it would be better if the test explicitly specified the diagnostic id to be used in call to `GetCodeActions` (perhaps in an overload).

omnisharp-dotnet/src/Services.UnitTests/DiagnosticWorker/QuickFixes/DiagnosticCodeActionsProviderTests.cs
```diff
@@ -0,0 +1,141 @@
...
+
+        private Diagnostic CreateDiagnostic() => Diagnostic.Create(Descriptor, null);
+
+        private Document CreateDocument()
+        {
+            var workspace = OmniSharpWorkspaceHelper.CreateOmnisharpWorkspaceWithDocument("test.cs", "");
+            return workspace.GetDocument("test.cs");
+        }
+
+        public static readonly DiagnosticDescriptor Descriptor = new(
```
duncanp-sonar, line 130: It's difficult to know when reading the tests whether the diagnostic id matters or not - in some tests it does, in others it doesn't.
Naming the descriptor differently might help e.g. `AnyDescriptor` if the actual id doesn't matter, or `WellKnownDescriptor` if it does.

omnisharp-dotnet/src/Services/DiagnosticWorker/QuickFixes/DiagnosticCodeActionsProvider.cs
duncanp-sonar: My brain hurts.

omnisharp-dotnet/src/Services/DiagnosticWorker/QuickFixes/DiagnosticCodeActionsProvider.cs
duncanp-sonar: It looks like it's a grouping mechanism i.e. a `CodeAction` might be a single action or a composite.

I think all of our current quick fixes are simple, so hopefully we don't need it.
Let's hope not, as I don't think it's public...

```
        /// <summary>
        /// Creates a <see cref="CodeAction"/> representing a group of code actions.
        /// </summary>
        /// <param name="title">Title of the <see cref="CodeAction"/> group.</param>
        /// <param name="nestedActions">The code actions within the group.</param>
        /// <param name="isInlinable"><see langword="true"/> to allow inlining the members of the group into the parent;
        /// otherwise, <see langword="false"/> to require that this group appear as a group with nested actions.</param>
        public static CodeAction Create(string title, ImmutableArray<CodeAction> nestedActions, bool isInlinable)
 ```

**Review by duncanp-sonar — COMMENTED** [review]
Looks good. Some comments about tests and the name of the new interface.

File: omnisharp-dotnet/src/Services/DiagnosticWorker/QuickFixes/DiagnosticCodeActionsProvider.cs [hunk 10]
```diff
@@ -0,0 +1,72 @@
...
+using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CodeActions;
+using Microsoft.CodeAnalysis.CodeFixes;
+using OmniSharp.Roslyn.CSharp.Services.Refactoring.V2;
+
+namespace SonarLint.OmniSharp.DotNet.Services.DiagnosticWorker.QuickFixes
+{
+    internal interface IDiagnosticCodeActionsProvider
```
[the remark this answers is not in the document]
rita-sonarsource: Regarding Diagnostic: there is already an OmniSharp interface `ICodeActionProvider`, which is poorly named and does more than providing code actions. Hence why I added the Diagnostic prefix. https://github.com/OmniSharp/omnisharp-roslyn/blob/ff21ac475cdf8afb15579f2b5fb346945db2a7e9/src/OmniSharp.Roslyn/Services/ICodeActionProvider.cs

omnisharp-dotnet/src/Services/DiagnosticWorker/QuickFixes/DiagnosticCodeActionsProvider.cs
rita-sonarsource: Thanks for checking!

PR #139, opened 2022-11-02
Add DiagnosticQuickFixesProvider

rita-sonarsource:


omnisharp-dotnet/src/Services/DiagnosticWorker/QuickFixes/DiagnosticQuickFixesProvider.cs
```diff
@@ -0,0 +1,138 @@
...
+            {
+                return Array.Empty<IQuickFix>();
+            }
+
+            var codeFixActions = await diagnosticCodeActionsProvider.GetCodeActions(diagnostic, document);
+            var quickFixes = new List<IQuickFix>();
+
+            foreach (var action in codeFixActions)
+            {
+                var fixes = new List<Fix>();
```
rita-sonarsource, line 100: Everything in the `foreach` loop is not tested yet

rita-sonarsource:
@duncanp-sonar please review, I will merge PR https://github.com/SonarSource/sonarlint-omnisharp/pull/141 into this one so the coverage will be higher + product code of the Operations will match what we discussed. I think it might be easier to review this before I merge.

omnisharp-dotnet/src/Services.UnitTests/DiagnosticWorker/QuickFixes/DiagnosticQuickFixesProviderTests.cs
```diff
@@ -0,0 +1,105 @@
...
+            Func<Task> act = async () => await testSubject.Handle(null);
+
+            await act.Should().ThrowExactlyAsync<NotSupportedException>();
+        }
+
+        [TestMethod]
+        public async Task GetDiagnosticQuickFixes_DocumentDoesNotExist_EmptyList()
+        {
+            var diagnostic = CreateDiagnostic();
+            var workspace = CreateOmniSharpWorkspace();
```
duncanp-sonar, line 54: This test is currently more a test of what happens with an empty workspace. Would it be better to create a workspace with a non-matching document, rather than a workspace with no documents?

duncanp-sonar (approved):
Just one minor suggestion about a test.

PR #141, opened 2022-11-04
tests added

ugras-ergun-sonarsource:


omnisharp-dotnet/src/Services/DiagnosticWorker/QuickFixes/IQuickFix.cs
```diff
@@ -25,13 +25,13 @@ namespace SonarLint.OmniSharp.DotNet.Services.DiagnosticWorker.QuickFixes
     public interface IQuickFix
     {
         string Message { get; }
-        IReadOnlyList<IFix> Fixes { get; }
+        IList<IFix> Fixes { get; }
```
ugras-ergun-sonarsource, line 28: I've changed these to Ilist to be able to add items.

omnisharp-dotnet/src/Services/DiagnosticWorker/QuickFixes/DiagnosticQuickFixesProvider.cs
```diff
@@ -48,16 +48,38 @@ public interface IDiagnosticQuickFixesProvider
     [Export(typeof(IDiagnosticQuickFixesProvider)), Shared]
     internal class DiagnosticQuickFixesProvider : BaseCodeActionService<IRequest, IAggregateResponse>, IDiagnosticQuickFixesProvider
     {
+        internal delegate Task<(Solution Solution, IEnumerable<FileOperationResponse> FileChanges)>
```
ugras-ergun-sonarsource, line 51: Because "GetFileChangesAsync" is in the base class and can not be mocked after having discussion with rita decided to add a delegate as a constructor parameter and pass to it in tests

omnisharp-dotnet/src/Services/DiagnosticWorker/QuickFixes/DiagnosticQuickFixesProvider.cs
```diff
@@ -48,16 +48,38 @@ public interface IDiagnosticQuickFixesProvider
...
             ILoggerFactory loggerFactory,
             OmniSharpOptions options,
-            ISonarLintDiagnosticCodeActionsProvider diagnosticCodeActionsProvider)
+            ISonarLintDiagnosticCodeActionsProvider diagnosticCodeActionsProvider) : this(
+            workspace,
+            codeActionProviders,
+            loggerFactory,
+            options,
+            diagnosticCodeActionsProvider,
+            null
```
ugras-ergun-sonarsource, line 71: GetFileChangesAsync is an instance method so the method wrapping should not be static and we can't pass non-static members. So I am doing a null check in the internal constructor to use the default value.

omnisharp-dotnet/src/Services.UnitTests/TestingInfrastructure/TestableCodeAction.cs
```diff
@@ -0,0 +1,45 @@
...
+ */
+
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.CodeAnalysis.CodeActions;
+
+namespace SonarLint.OmniSharp.DotNet.Services.UnitTests.TestingInfrastructure
+{
+    internal class TestableCodeAction : CodeAction
```
ugras-ergun-sonarsource, line 28: Because code action is an abstract class we can not mock it so I've created a mock class.

omnisharp-dotnet/src/Services.UnitTests/TestingInfrastructure/TestableCodeAction.cs
```diff
@@ -0,0 +1,45 @@
...
+
+        internal TestableCodeAction(string title, IEnumerable<CodeActionOperation> operations)
+        {
+            Title = title;
+            this.operations = operations;
+        }
+
+        public override string Title { get; }
+
+        protected override async Task<IEnumerable<CodeActionOperation>> ComputeOperationsAsync(CancellationToken cancellationToken)
```
ugras-ergun-sonarsource, line 40: This is to "GetOperationsAsync" but unfortunately I had to go down to this method which was virtual and did not have any internal parameters.

omnisharp-dotnet/src/Services/DiagnosticWorker/QuickFixes/IQuickFix.cs
rita-sonarsource: The whole point of the readonly was so you could not add items. Why not just mock it to return whatever you need?

ugras-ergun-sonarsource: Yeah and apparently I've changed the code where I needed to add things.

omnisharp-dotnet/src/Services/DiagnosticWorker/QuickFixes/DiagnosticQuickFixesProvider.cs
```diff
@@ -48,16 +48,38 @@ public interface IDiagnosticQuickFixesProvider
     [Export(typeof(IDiagnosticQuickFixesProvider)), Shared]
     internal class DiagnosticQuickFixesProvider : BaseCodeActionService<IRequest, IAggregateResponse>, IDiagnosticQuickFixesProvider
     {
+        internal delegate Task<(Solution Solution, IEnumerable<FileOperationResponse> FileChanges)>
+            GetFileChangesAsyncFnc(Solution newSolution, Solution oldSolution, string directory, bool wantTextChanges,
+                bool wantsAllCodeActionOperations);
+
         private readonly ISonarLintDiagnosticCodeActionsProvider diagnosticCodeActionsProvider;
         private readonly OmniSharpWorkspace workspace;
+        private readonly GetFileChangesAsyncFnc getFileChangesAsyncFnc;
```
rita-sonarsource, line 57: cosmetics: maybe with a "u" in Func? I believe that's the standard

omnisharp-dotnet/src/Services.UnitTests/TestingInfrastructure/TestableCodeAction.cs
rita-sonarsource, line 28: I think it should be as a private class next to the code that needs it. We shouldn't reference CodeAction object anywhere else because it's so complicated.

omnisharp-dotnet/src/Services.UnitTests/DiagnosticWorker/QuickFixes/DiagnosticQuickFixesProviderTests.cs
```diff
@@ -85,9 +88,90 @@ public async Task GetDiagnosticQuickFixes_DiagnosticHasNoCodeActions_EmptyList()
...
         }
 
+        [TestMethod]
+        public async Task GetDiagnosticQuickFixes_DiagnosticHasCodeActions_ReturnsQuickFixes()
+        {
+            var diagnostic = CreateDiagnostic();
+            var workspace = CreateOmnisharpWorkspaceWithDocument("test.cs", "content");
+            var document = workspace.GetDocument("test.cs");
+
+            var operation = new ApplyChangesOperation(workspace.CurrentSolution);
```
rita-sonarsource, line 98: `operation` is not used anywhere besides passing it to CreateCodeAction: can we simplify CreateCodeAction to receive the solution and do the orchestration inside? it would simplify reading

rita-sonarsource, line 98: Perhaps also a check what happens if the operation is not ApplyChangesOperation?

omnisharp-dotnet/src/Services.UnitTests/DiagnosticWorker/QuickFixes/DiagnosticQuickFixesProviderTests.cs
```diff
@@ -85,9 +88,90 @@ public async Task GetDiagnosticQuickFixes_DiagnosticHasNoCodeActions_EmptyList()
...
+
+            var codeAction = CreateCodeAction("fix message", operation);
+
+            var diagnosticCodeActionsProvider =
+                CreateDiagnosticCodeActionsProvider(diagnostic, document, new List<CodeAction> { codeAction });
+
+            var textChange = CreateTextChange("New Text", 1, 2, 3, 4);
```
rita-sonarsource, line 105: it would help to add the parameter names

omnisharp-dotnet/src/Services/DiagnosticWorker/QuickFixes/DiagnosticQuickFixesProvider.cs
```diff
@@ -97,15 +120,15 @@ public async Task<IQuickFix[]> GetDiagnosticQuickFixes(Diagnostic diagnostic, st
 
             foreach (var action in codeFixActions)
             {
-                var fixes = new List<Fix>();
+                var fixes = new List<IFix>();
                 var solution = workspace.CurrentSolution;
                 var directory = Path.GetDirectoryName(filePath);
                 var operations = await action.GetOperationsAsync(CancellationToken.None);
 
                 foreach (var operation in operations.OfType<ApplyChangesOperation>())
```
rita-sonarsource, line 128: We have no unit tests of multiple ApplyChangesOperation. 
Perhaps we should instead do SingleOrDefault and assert that it's not null?

omnisharp-dotnet/src/Services/DiagnosticWorker/QuickFixes/DiagnosticQuickFixesProvider.cs
```diff
@@ -134,5 +157,13 @@ Edit ToEdit(LinePositionSpanTextChange textChange)
                     endColumn: textChange.EndColumn);
             }
         }
+        private GetFileChangesAsyncFnc GetFileChangesAsyncFncImpl()
```
rita-sonarsource, line 160: cosmetics: missing empty line

omnisharp-dotnet/src/Services.UnitTests/DiagnosticWorker/QuickFixes/DiagnosticQuickFixesProviderTests.cs
```diff
@@ -85,9 +88,90 @@ public async Task GetDiagnosticQuickFixes_DiagnosticHasNoCodeActions_EmptyList()
...
+            result[0].Fixes[0].FileName.Should().Be("test.cs");
+            result[0].Fixes[0].Edits.Count.Should().Be(1);
+            result[0].Fixes[0].Edits[0].NewText.Should().Be("New Text");
+            result[0].Fixes[0].Edits[0].StartColumn.Should().Be(1);
+            result[0].Fixes[0].Edits[0].EndColumn.Should().Be(2);
+            result[0].Fixes[0].Edits[0].StartLine.Should().Be(3);
+            result[0].Fixes[0].Edits[0].EndLine.Should().Be(4);
+        }
+
+
```
rita-sonarsource, line 128: cosmetics: extra empty line

omnisharp-dotnet/src/Services.UnitTests/DiagnosticWorker/QuickFixes/DiagnosticQuickFixesProviderTests.cs
```diff
@@ -85,9 +88,90 @@ public async Task GetDiagnosticQuickFixes_DiagnosticHasNoCodeActions_EmptyList()
...
+            result[0].Fixes[0].Edits[0].EndColumn.Should().Be(2);
+            result[0].Fixes[0].Edits[0].StartLine.Should().Be(3);
+            result[0].Fixes[0].Edits[0].EndLine.Should().Be(4);
+            result[0].Fixes[0].Edits[1].NewText.Should().Be("New Text 2");
+            result[0].Fixes[0].Edits[1].StartColumn.Should().Be(5);
+            result[0].Fixes[0].Edits[1].EndColumn.Should().Be(6);
+            result[0].Fixes[0].Edits[1].StartLine.Should().Be(7);
+            result[0].Fixes[0].Edits[1].EndLine.Should().Be(8);
+        }
         private DiagnosticQuickFixesProvider CreateTestSubject(
```
rita-sonarsource, line 171: cosmetics: missing empty line

omnisharp-dotnet/src/Services.UnitTests/DiagnosticWorker/QuickFixes/DiagnosticQuickFixesProviderTests.cs
```diff
@@ -99,7 +183,51 @@ public async Task GetDiagnosticQuickFixes_DiagnosticHasNoCodeActions_EmptyList()
...
+            {
+                NewText = newText,
+                StartColumn = startColumn,
+                EndColumn = endColumn,
+                StartLine = startLine,
+                EndLine = endLine
+            };
+            return textChange;
+        }
+        private Mock<ISonarLintDiagnosticCodeActionsProvider> CreateDiagnosticCodeActionsProvider(
```
rita-sonarsource, line 210: cosmetics: missing empty line

omnisharp-dotnet/src/Services.UnitTests/DiagnosticWorker/QuickFixes/DiagnosticQuickFixesProviderTests.cs
```diff
@@ -99,7 +183,51 @@ public async Task GetDiagnosticQuickFixes_DiagnosticHasNoCodeActions_EmptyList()
...
+            List<CodeAction> codeActions = null)
+        {
+            codeActions ??= new List<CodeAction>();
+
+            var diagnosticCodeActionsProvider = new Mock<ISonarLintDiagnosticCodeActionsProvider>();
+            diagnosticCodeActionsProvider.Setup(dap => dap.GetCodeActions(diagnostic, document))
+                .ReturnsAsync(codeActions);
+            return diagnosticCodeActionsProvider;
+        }
+        private TestableCodeAction CreateCodeAction(string title, ApplyChangesOperation operation)
```
rita-sonarsource, line 221: cosmetics: missing empty line

omnisharp-dotnet/src/Services.UnitTests/DiagnosticWorker/QuickFixes/DiagnosticQuickFixesProviderTests.cs
```diff
@@ -99,7 +183,51 @@ public async Task GetDiagnosticQuickFixes_DiagnosticHasNoCodeActions_EmptyList()
...
+        {
+            return new TestableCodeAction(title, new List<CodeActionOperation> { operation });
+        }
+        private static ModifiedFileResponse CreateFileResponse(string fileName,
```
rita-sonarsource, line 225: cosmetics: missing empty line

omnisharp-dotnet/src/Services.UnitTests/DiagnosticWorker/QuickFixes/DiagnosticQuickFixesProviderTests.cs
```diff
@@ -99,7 +183,51 @@ public async Task GetDiagnosticQuickFixes_DiagnosticHasNoCodeActions_EmptyList()
                 codeActionProviders,
                 Mock.Of<ILoggerFactory>(),
                 new OmniSharpOptions(),
-                diagnosticCodeActionsProvider);
+                diagnosticCodeActionsProvider,
+                mockFunction);
+        }
+
+        private DiagnosticQuickFixesProvider.GetFileChangesAsyncFnc GetFileChangesAsyncMock(Solution solution,
```
rita-sonarsource, line 190: This should be a mock + verify that it was called with the right params

omnisharp-dotnet/src/Services.UnitTests/DiagnosticWorker/QuickFixes/DiagnosticQuickFixesProviderTests.cs
```diff
@@ -99,7 +183,51 @@ public async Task GetDiagnosticQuickFixes_DiagnosticHasNoCodeActions_EmptyList()
...
+        private Mock<ISonarLintDiagnosticCodeActionsProvider> CreateDiagnosticCodeActionsProvider(
+            Diagnostic diagnostic, Document document,
+            List<CodeAction> codeActions = null)
+        {
+            codeActions ??= new List<CodeAction>();
+
+            var diagnosticCodeActionsProvider = new Mock<ISonarLintDiagnosticCodeActionsProvider>();
+            diagnosticCodeActionsProvider.Setup(dap => dap.GetCodeActions(diagnostic, document))
+                .ReturnsAsync(codeActions);
+            return diagnosticCodeActionsProvider;
```
rita-sonarsource, line 219: Cosmetics: new line before return

omnisharp-dotnet/src/Services.UnitTests/DiagnosticWorker/QuickFixes/DiagnosticQuickFixesProviderTests.cs
```diff
@@ -99,7 +183,51 @@ public async Task GetDiagnosticQuickFixes_DiagnosticHasNoCodeActions_EmptyList()
...
+        {
+            var textChange = new LinePositionSpanTextChange
+            {
+                NewText = newText,
+                StartColumn = startColumn,
+                EndColumn = endColumn,
+                StartLine = startLine,
+                EndLine = endLine
+            };
+            return textChange;
```
rita-sonarsource, line 208: Cosmetics: new line before return

omnisharp-dotnet/src/Services.UnitTests/DiagnosticWorker/QuickFixes/DiagnosticQuickFixesProviderTests.cs
```diff
@@ -99,7 +183,51 @@ public async Task GetDiagnosticQuickFixes_DiagnosticHasNoCodeActions_EmptyList()
...
+        private TestableCodeAction CreateCodeAction(string title, ApplyChangesOperation operation)
+        {
+            return new TestableCodeAction(title, new List<CodeActionOperation> { operation });
+        }
+        private static ModifiedFileResponse CreateFileResponse(string fileName,
+            params LinePositionSpanTextChange[] changes)
+        {
+            var fileResponse = new ModifiedFileResponse("test.cs");
+            fileResponse.Changes = changes;
+            return fileResponse;
```
rita-sonarsource, line 230: Cosmetics: new line before return

omnisharp-dotnet/src/Services/DiagnosticWorker/QuickFixes/DiagnosticQuickFixesProvider.cs
duncanp-sonar: FWIW I don't normally have "Func" on the end - it's clear from the method signature. I'd just call it `GetFileChangesAsync`.

rita-sonarsource (changes requested):
@ugras-ergun-sonarsource looks good, see my comments (and my cosmetic comments 😜 )

omnisharp-dotnet/src/Services/DiagnosticWorker/QuickFixes/DiagnosticQuickFixesProvider.cs
```diff
@@ -67,6 +89,7 @@ internal class DiagnosticQuickFixesProvider : BaseCodeActionService<IRequest, IA
         {
             this.workspace = workspace;
             this.diagnosticCodeActionsProvider = diagnosticCodeActionsProvider;
+            this.getFileChangesAsyncFnc = getFileChangesAsyncFnc ?? GetFileChangesAsyncFncImpl();
```
duncanp-sonar, line 92: #Can't the constructor just reference the `GetFileChangesAsync` method directly? e.g.

```
//this.getFileChangesAsyncFnc = getFileChangesAsyncFnc ?? GetFileChangesAsyncFncImpl();

this.getFileChangesAsyncFnc = getFileChangesAsyncFnc ?? GetFileChangesAsync;
```

omnisharp-dotnet/src/Services/DiagnosticWorker/QuickFixes/DiagnosticQuickFixesProvider.cs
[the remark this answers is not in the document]
duncanp-sonar: (...although the method might not be needed)

omnisharp-dotnet/src/Services/DiagnosticWorker/QuickFixes/DiagnosticQuickFixesProvider.cs
duncanp-sonar: I think it's worth commenting explicitly that this delegate was introduced to allow tests to replace the behaviour of the base class - it will save some time for anyone who has to look at this in the future.

omnisharp-dotnet/src/Services/DiagnosticWorker/QuickFixes/DiagnosticQuickFixesProvider.cs
duncanp-sonar: As discussed offline, I think there's a preceding problem, namely that we filter out any changes that are not `ApplyChangesOperation`. Presumably all of the operations in the code action need to be executed for it to make sense, not just some of them.

If we only support simple edits to the current file currently, we could remove the filter **and** @rita-sonarsource's suggestion i.e. we'll only apply the fix if there is only one operation and it is of type `ApplyChangesOperation`.

(alternatively, if we think we can handle multiple `ApplyChangesOperation`s then we could include the fix if all of its operations are of type `ApplyChangeOperation`, but then we'd need a test for it as @rita-sonarsource suggests).

omnisharp-dotnet/src/Services.UnitTests/DiagnosticWorker/QuickFixes/DiagnosticQuickFixesProviderTests.cs
```diff
@@ -85,9 +88,90 @@ public async Task GetDiagnosticQuickFixes_DiagnosticHasNoCodeActions_EmptyList()
...
+            result[0].Fixes[0].Edits[0].NewText.Should().Be("New Text");
+            result[0].Fixes[0].Edits[0].StartColumn.Should().Be(1);
+            result[0].Fixes[0].Edits[0].EndColumn.Should().Be(2);
+            result[0].Fixes[0].Edits[0].StartLine.Should().Be(3);
+            result[0].Fixes[0].Edits[0].EndLine.Should().Be(4);
+        }
+
+
+        [TestMethod]
+        public async Task GetDiagnosticQuickFixes_DiagnosticHasCodeActionswithMultipleEdits_ReturnsQuickFixes()
```
duncanp-sonar, line 130: Naming: this test only has one `CodeAction` so it should be `..._DiagnosticHasCodeActionWithMultipleEdits...`.

duncanp-sonar: Missing test case: a diagnostic with two different code actions.
Ideally the new test would check that the first code fix in each action is given the same solution instance i.e. a "clean" solution is used each time.

omnisharp-dotnet/src/Services/DiagnosticWorker/QuickFixes/DiagnosticQuickFixesProvider.cs
ugras-ergun-sonarsource: GetFileChangesAsync is the name of the method in the base class and if we name it that way it would create a conflict.

omnisharp-dotnet/src/Services/DiagnosticWorker/QuickFixes/DiagnosticQuickFixesProvider.cs
```diff
@@ -97,23 +120,24 @@ public async Task<IQuickFix[]> GetDiagnosticQuickFixes(Diagnostic diagnostic, st
 
             foreach (var action in codeFixActions)
             {
-                var fixes = new List<Fix>();
-                var solution = workspace.CurrentSolution;
```
ugras-ergun-sonarsource, line 101: after the changes, there was no need to have variables so I have simplified the code.

omnisharp-dotnet/src/Services/DiagnosticWorker/QuickFixes/DiagnosticQuickFixesProvider.cs
```diff
@@ -97,23 +120,24 @@ public async Task<IQuickFix[]> GetDiagnosticQuickFixes(Diagnostic diagnostic, st
...
                 var directory = Path.GetDirectoryName(filePath);
                 var operations = await action.GetOperationsAsync(CancellationToken.None);
 
-                foreach (var operation in operations.OfType<ApplyChangesOperation>())
-                {
-                    var solutionAfterOperation = operation.ChangedSolution;
-                    var fileChangesResult = await GetFileChangesAsync(solutionAfterOperation, solution, directory, true, false);
+                var applyChangesOperations = operations.OfType<ApplyChangesOperation>().ToList();
+                
+                Debug.Assert(applyChangesOperations.Count <= 1);
```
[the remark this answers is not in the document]
duncanp-sonar: It's always better to include a message in an assertion. I you don't have a debugger attached, the dialogue that pops up isn't that helpful, as it just shows the call stack.

omnisharp-dotnet/src/Services/DiagnosticWorker/QuickFixes/DiagnosticQuickFixesProvider.cs
```diff
@@ -97,23 +120,24 @@ public async Task<IQuickFix[]> GetDiagnosticQuickFixes(Diagnostic diagnostic, st
...
-                    Debug.Assert(fileChangesResult.FileChanges.All(c => c is ModifiedFileResponse));
+                if (applyChangesOperations.Count == 0)
+                {
+                    continue;
+                }
+                var fileChangesResult = await getFileChangesAsyncFunc(applyChangesOperations.Single().ChangedSolution, workspace.CurrentSolution, directory, true, false);
 
-                    var fileFixes = fileChangesResult.FileChanges.Select(c => ToFix((ModifiedFileResponse)c));
-                    fixes.AddRange(fileFixes);
+                Debug.Assert(fileChangesResult.FileChanges.All(c => c is ModifiedFileResponse));
```
duncanp-sonar, line 137: Missing assertion message.

omnisharp-dotnet/src/Services/DiagnosticWorker/QuickFixes/DiagnosticQuickFixesProvider.cs
```diff
@@ -97,23 +120,24 @@ public async Task<IQuickFix[]> GetDiagnosticQuickFixes(Diagnostic diagnostic, st
...
-                var solution = workspace.CurrentSolution;
+                var fixes = new List<IFix>();
                 var directory = Path.GetDirectoryName(filePath);
                 var operations = await action.GetOperationsAsync(CancellationToken.None);
 
-                foreach (var operation in operations.OfType<ApplyChangesOperation>())
-                {
-                    var solutionAfterOperation = operation.ChangedSolution;
-                    var fileChangesResult = await GetFileChangesAsync(solutionAfterOperation, solution, directory, true, false);
+                var applyChangesOperations = operations.OfType<ApplyChangesOperation>().ToList();
```
duncanp-sonar, line 127: This logic has changed back from what @rita-sonarsource [shared](https://github.com/SonarSource/sonarlint-omnisharp/pull/136/commits/be5fc23122c1b182a027c0cac14f70b4dcd791f6) this morning - it's now back to the more-unsafe version of ignoring any non-`ApplyChangesOperations`.

The safer approach is the one in Rita's version: only apply the fix if there is only one operation, and that operation is of type `ApplyChangesOperation`.

rita-sonarsource:
@duncanp-sonar please re-review

omnisharp-dotnet/src/Services/DiagnosticWorker/QuickFixes/DiagnosticQuickFixesProvider.cs
```diff
@@ -97,23 +123,28 @@ public async Task<IQuickFix[]> GetDiagnosticQuickFixes(Diagnostic diagnostic, st
...
+                var fixes = new List<IFix>();
                 var directory = Path.GetDirectoryName(filePath);
                 var operations = await action.GetOperationsAsync(CancellationToken.None);
 
-                foreach (var operation in operations.OfType<ApplyChangesOperation>())
+                if (operations.Length > 1)
                 {
-                    var solutionAfterOperation = operation.ChangedSolution;
-                    var fileChangesResult = await GetFileChangesAsync(solutionAfterOperation, solution, directory, true, false);
+                    throw new ArgumentOutOfRangeException(nameof(operations), "Expecting quick fixes to contain one or zero operations.");
```
duncanp-sonar, line 132: Throwing is pretty strong - throwing here will mean no diagnostics or fixes for the current file (?)

Your [version from yesterday](https://github.com/SonarSource/sonarlint-omnisharp/pull/136/commits/be5fc23122c1b182a027c0cac14f70b4dcd791f6) just skipped that particular fix.

Would a `Debug.Assert` be enough?

omnisharp-dotnet/src/Services/DiagnosticWorker/QuickFixes/DiagnosticQuickFixesProvider.cs
duncanp-sonar: The dialogue popped up by VS is more useful if the assertion has a message.

duncanp-sonar (approved):
Looks good - just the one question about whether throwing is too strong.

omnisharp-dotnet/src/Services/DiagnosticWorker/QuickFixes/DiagnosticQuickFixesProvider.cs
rita-sonarsource: I'm not sure. If it's a debug.assert, we won't likely ever see it in production and it won't work properly. I think it's better to leave the exception so that we would at least know if we're wrong in our assumption.

rita-sonarsource:
@duncanp-sonar I will merge this with the exception throwing for now, so I could start working on the conflicts of the other PRs. I will fix this to debug assert later if we decide to do it.

omnisharp-dotnet/src/Services/DiagnosticWorker/QuickFixes/DiagnosticQuickFixesProvider.cs
duncanp-sonar: An exception in production makes the experience worse for users (no diagnostics for the whole file), but there's no guarantee it would be reported. I'm leaning towards impacting the user as little as possible e.g. assert, and log a message, so that only "incompatible" fixes are ignored.

PR #153, opened 2023-05-22
Add an IT to reproduce FP with dotnet 7 projects

henryju:
See https://github.com/SonarSource/sonar-dotnet/issues/7230


thahnen (approved):
LGTM to me, I'd merge that before the SLI-957 PR and rerun it once as I'm not sure if the new test will also work on the new version, this is up to you! 😄

PR #154, opened 2023-05-30
SLI-942: Update C#/VB.NET analyzer

thahnen:


thahnen:
The commits will be merged, therefore only referencing the SLI-962 (and not the wrong ticket in the second to last commit)

thahnen:
Resolved #155

PR #158, opened 2023-06-22
update to v1.39.6

ugras-ergun-sonarsource:


its/pom.xml
```diff
@@ -93,7 +92,7 @@
               <goal>wget</goal>
             </goals>
             <configuration>
-              <url>https://repox.jfrog.io/artifactory/sonarsource/org/sonarsource/sonarlint/omnisharp/omnisharp-roslyn/1.39.0/omnisharp-roslyn-1.39.0-net472.zip</url>
+              <url>https://repox.jfrog.io/artifactory/sonarsource/org/sonarsource/sonarlint/omnisharp/omnisharp-roslyn/1.39.6/omnisharp-roslyn-1.39.6-net472.zip</url>
```
rita-gorokhod, line 95: why not use a build property?

omnisharp-dotnet/src/Services/SonarLint.OmniSharp.DotNet.Services.csproj
```diff
@@ -13,12 +13,12 @@
          Referencing a different version of NewtonSoft.Json doesn't seem to cause a problem,
```
rita-gorokhod, line 13: the newtonsoft comment is irrelevant now, we are _not_ targetting the same version (it was changed to 13.0.3 in their release)

rita-gorokhod:
@ugras-ergun-sonarsource LGTM, just see my 2 comments

its/pom.xml
ugras-ergun-sonarsource: Git thinks this is a change but this is already in the master, handled by @henryju

thahnen:
@henryju regarding the flaky tests, should we also add the [awaitility](http://www.awaitility.org/) library in the coming future?

its/pom.xml
henryju: FYI having plain URL is simpler when we want to test a different distribution (for example use the official release from OmniSharp).

PR #163, opened 2023-08-11
Use v2.0.1 of pipelines-yaml-templates

duncanp-sonar:
i.e. get the updated templates that fix the Mend agent error

Fixes https://sonarsource.atlassian.net/browse/SLI-1056

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -12,7 +12,7 @@ resources:
     - repository: commonTemplates
       type: git
       name: pipelines-yaml-templates
-      ref:  refs/tags/v1.0.17
+      ref:  refs/tags/v2.0.1
 
 
 stages:
@@ -90,10 +90,10 @@ stages:
               
         - task: SonarQubeAnalyze@5
         
-        - template: whitesource.yml@commonTemplates
+        - template: mend.yml@commonTemplates
           parameters:
             productName : 'SonarLint/OmniSharp'
-            # We're running two WhiteSource scans in this pipeline, one for .NET and one for Java.
+            # We're running two Mend scans in this pipeline, one for .NET and one for Java.
             # They can't both have the same projectName. We'll use the default projectName for the Java
             # project so it works with the releasability check.
             projectName : 'SonarSource/sonarlint-omnisharp/dotnet'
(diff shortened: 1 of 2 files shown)


azure-pipelines.yml
```diff
@@ -12,7 +12,7 @@ resources:
     - repository: commonTemplates
       type: git
       name: pipelines-yaml-templates
-      ref:  refs/tags/v1.0.17
```
duncanp-sonar, line 15: The fix is actually this commit in the `pipelines-yaml-templates` repo [here](https://dev.azure.com/sonarsource/DotNetTeam%20Project/_git/pipelines-yaml-templates/commit/088d57f8e41c1cd2c0ec1a49c9af84ce50fb47ca?refName=refs%2Fheads%2Fmaster). That commit changes the pipeline to use the same [package feed](https://dev.azure.com/sonarsource/DotNetTeam%20Project/_artifacts/feed/ExternalTools) as SLVS, which contains the up to date Mend agent.

azure-pipelines.yml
```diff
@@ -90,10 +90,10 @@ stages:
               
         - task: SonarQubeAnalyze@5
         
-        - template: whitesource.yml@commonTemplates
+        - template: mend.yml@commonTemplates
```
duncanp-sonar, line 93: Previously this pipeline was importing the version of the `pipelines-yaml-templates` tagged `v1.0.17`.

There is a newer `v2.0.0` [tag](https://dev.azure.com/sonarsource/DotNetTeam%20Project/_git/pipelines-yaml-templates/tags) which doesn't contain any functional changes. However, it renames `Whitespace` to `Mend` everywhere - comments, script names etc. The rest of the changes in this pipeline are to make it work with those renamings.

duncanp-sonar:
Previously the execution of Mend was failing;

image


Now, both Mend steps are working (one for the .NET code and one for the Java code). See e.g. [this build](https://dev.azure.com/sonarsource/DotNetTeam%20Project/_build/results?buildId=75931&view=logs&s=c11308f6-c6d2-50c8-c26c-9862ac5faec9):

#### .NET
image


#### Java
image

Note: the Mend analysis normally only run in CI builds for `master` (which is why we didn't notice this error before).
That also means the Mend tasks are skipped in the CI build triggered for this PR. However, you can force it to run by  setting the variable `ForceMendScan` to `true` when manually triggering a build:

image

its/src/test/java/org/sonarsource/sonarlint/omnisharp/its/OmnisharpIntegrationTests.java
```diff
@@ -299,9 +299,9 @@ void analyzeMixedSolutionWithOldOmnisharp(@TempDir Path tmpDir) throws Exception
         tuple("csharpsquid:S1135", "Complete the task associated to this 'TODO' comment.", 1, 3, 1, 7, inputFileFramework.getPath(), INFO));
   }
 
-  @Test
-  // FIXME
-  @DisabledOnOs(OS.WINDOWS)
```
duncanp-sonar, line 304: This is a separate issue from the Mend failure.
This test was previously failing only on Windows, but it's now failing on all three platforms.

thahnen (approved):
LGTM! The issue with the failing test will be part of the SLI hardening / analyzer update next week or the week after!

PR #192, opened 2024-01-25
NO-JIRA: Use channel for all Java develoeprs

thahnen:
This is only a quality of life change: Since VSCode squad now also uses SLOMNI for SLVSCODE, inform them on releases automatically as well.


thahnen:
Closed as there won't be a `team-sonarlint-java` Slack channel anymore.

PR #213, opened 2024-07-29
SLI-1511 Update embedded C# analyzer to 9.30.0.95878

nquinquenel:


pom.xml
```diff
@@ -42,7 +42,7 @@
     <gitRepositoryName>sonarlint-omnisharp</gitRepositoryName>
     <artifactsToPublish>${project.groupId}:sonarlint-omnisharp-plugin:jar</artifactsToPublish>
     <junit.version>5.9.1</junit.version>
-    <sonarlint-core.version>9.0.0.74282</sonarlint-core.version>
```
nquinquenel, line 45: Updated SLCORE because why not, but only to the latest non-SLOOP version. It requires a lot of refactoring for the integration tests otherwise.

thahnen (approved):
LGTM

PR #217, opened 2024-08-12
SLOMNI-12 Use nuget proxy from repox

eray-felek-sonarsource:


jblievremont (commented):
Nice! I would also update the build instructions in the README to mention the newly required environment variables.

jblievremont (approved):

PR #222, opened 2024-08-15
BUILD-5998 Compute version for SonarQube analysis of .NET code

jblievremont:


.cirrus.yaml
```diff
@@ -35,6 +35,16 @@ build_dotnet_task:
...
     cpu: 4
     memory: 4G
-  build_script:
-    - source cirrus-env BUILD
-    # - regular_mvn_build_deploy_analyze -Dmaven.test.skip=true -Dsonar.skip=true -pl '!java-checks-test-sources/default,!java-checks-test-sources/aws'
+  build_script: |
+    source cirrus-env BUILD
+
+    # Calculate the POM file path
+    versionFilePath="${CIRRUS_WORKING_DIR}/pom.xml"
```
thahnen, line 42: Why not use `CURRENT_VERSION=$(maven_expression "project.version")` instead of adding a dependency? This `maven_expression` is built into our Cirrus CI scripts, as far as I know.

jayadeep-km-sonarsource: can confirm that `maven_expression` is part of common cirrusci scripts

jblievremont: Good catch! We blindly ported the script from Azure 🙈

jayadeep-km-sonarsource (approved):

PR #227, opened 2024-08-19
BUILD-6029: Validate stage

jayadeep-km-sonarsource:
BUILD-6029: Validate stage

### DIFF
--- a/.cirrus.yaml
+++ b/.cirrus.yaml
@@ -158,3 +158,30 @@ qa_windows_task:
       path: '**/surefire-reports/TEST-*.xml'
   cleanup_before_cache_script: cleanup_maven_repository
   <<: *NOTIFY_SLACK_ON_FAILURE
+
+validate_windows_task:
+  depends_on:
+    - build_dotnet
+  ec2_instance:
+    <<: *INSTANCE_DEFINITION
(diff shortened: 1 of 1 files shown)


.cirrus.yaml
```diff
@@ -109,3 +115,30 @@ build_task:
...
   cleanup_before_cache_script: cleanup_maven_repository
+
+validate_windows_task:
+  depends_on:
+    - build_dotnet
+  ec2_instance:
+    <<: *INSTANCE_DEFINITION
+  get_jarcontentfiles_script: |
+    curl -O "https://api.cirrus-ci.com/v1/artifact/build/${CIRRUS_BUILD_ID}/build_dotnet/jarcontentfiles.zip"
+    unzip jarcontentfiles.zip
```
jblievremont, line 126: Nitpick: maybe we can extract this to a variable (also used in the Maven build and `validate_linux` stages)

jblievremont (approved):
See my small suggestion, otherwise LGTM 👍🏻

.cirrus.yaml
jayadeep-km-sonarsource: Will address as part of cleanup in a separate PR

PR #274, opened 2025-09-03
SLOMNI-63 Exclude razor files from analysis

gabriela-trutan-sonarsource:
Razor files are not well supported, and raise a lot of false-positives. We should exclude them for now.


damien-urruty-sonarsource (approved):
LGTM for the Java side. I don't know for the C# part but it's all about test resources so not a big deal

PR #301, opened 2026-02-17
SLOMNI-101 Try removing unused dependency

eray-felek-sonarsource:


voz:
@eray-felek-sonarsource  can we close it or define path forward if we need to? Thx for helping to keep things tidy.

PR #333, opened 2026-06-26
SLOMNI-128 Use Sdk: configuration for net6 OmniSharp

nquinquenel:
## Summary

- Pass `Sdk:Path` and `Sdk:Version` when starting the net6 OmniSharp process instead of `MsBuild:MSBuildOverride:MSBuildPath`, which is only valid for the legacy Framework OmniSharp flavor.
- Add `DotNetSdkPathResolver` to derive SDK path and version from the MSBuild/SDK path hint provided by IDEs (e.g. Rider's `activeMsBuildPath`).
- Keep `MSBuildOverride` for the legacy mono/Windows OmniSharp build path, and as a fallback when the hint is not an SDK directory.

SDK version selection (e.g. avoiding .NET 10 when the project targets an older SDK) is intentionally out of scope — the IDE hint is passed through as-is. Follow-up work can address auto-selection if needed.

## Test plan

- [x] `mvn -pl omnisharp-plugin test -Dtest=OmnisharpCommandBuilderTests,DotNetSdkPathResolverTests`
- [x] `mvn -pl omnisharp-plugin test`
- [ ] Verify Rider analysis uses the SDK path from the IDE hint on Linux


nquinquenel:
gitar auto-apply:on

its/src/test/projects/ConsoleApp1WithTestProject/ConsoleApp1/ConsoleApp1.csproj
```diff
@@ -2,7 +2,7 @@
 
     <PropertyGroup>
         <OutputType>Exe</OutputType>
-        <TargetFramework>net5.0</TargetFramework>
```
damien-urruty-sonarsource, line 5: Does it mean we don't test support of net5.0 anymore? Is this a problem?

.github/workflows/build.yml
```diff
@@ -201,6 +201,13 @@ jobs:
             development/artifactory/token/{REPO_OWNER_NAME_DASH}-private-reader username | ARTIFACTORY_USERNAME;
             development/artifactory/token/{REPO_OWNER_NAME_DASH}-private-reader access_token | ARTIFACTORY_ACCESS_TOKEN;
 
+      - name: Install .NET SDKs
+        uses: actions/setup-dotnet@v4
```
damien-urruty-sonarsource: Are we allowed to pull dependencies like this? Shouldn't we rely on Mise?

omnisharp-plugin/src/main/java/org/sonarsource/sonarlint/omnisharp/DotNetSdkPathResolver.java
```diff
@@ -0,0 +1,86 @@
...
+        if (fileName != null && isSdkVersion(fileName.toString())) {
+          return Optional.of(new SdkConfiguration(current, fileName.toString()));
+        }
+      }
+      current = parent;
+    }
+    return Optional.empty();
+  }
+
+  private static boolean isSdkDirectory(Path parent) {
```
damien-urruty-sonarsource: Shouldn't the parameter be named `directory` in this case?

omnisharp-plugin/src/main/java/org/sonarsource/sonarlint/omnisharp/OmnisharpCommandBuilder.java
```diff
@@ -79,6 +79,23 @@ public ProcessBuilder build(Path projectBaseDir, @Nullable Path monoPath, @Nulla
     return addArguments(projectBaseDir, msBuildPath, solutionPath, loadProjectsOnDemand, args);
   }
 
+  private ProcessBuilder addNet6Arguments(Path projectBaseDir, @Nullable Path msBuildPath, @Nullable Path solutionPath, boolean loadProjectsOnDemand, List<String> args) {
+    args.add("-v");
+    if (sonarLintRuntime.getClientPid() != 0) {
+      args.add("--hostPID");
+      args.add(Long.toString(sonarLintRuntime.getClientPid()));
+    }
```
damien-urruty-sonarsource: Shouldn't those be part of the common arguments? I can see the same in addArguments (which we should probably rename by the way?)

damien-urruty-sonarsource (commented):
Looks mostly good to me, some comments to improve the code.
I think it lacks some integration tests to verify the behavior end-to-end. THat would help validate the change of behavior

its/src/test/projects/ConsoleApp1WithTestProject/ConsoleApp1/ConsoleApp1.csproj
nquinquenel: It's still supported by OmniSharp, we just don't test against it anymore 🤷🏻

.github/workflows/build.yml
nquinquenel: I think this is standard way of installing dotnet. There's no mise for dotnet in this project.
- Linux relies on `apt-get install dotnet-sdk-8.0`
- Windows relies on `setup-dotnet`
Do you know a way to rely on mise for all dotnet dependencies?

omnisharp-plugin/src/main/java/org/sonarsource/sonarlint/omnisharp/DotNetSdkPathResolver.java
nquinquenel: Agree, makes sense

omnisharp-plugin/src/main/java/org/sonarsource/sonarlint/omnisharp/OmnisharpCommandBuilder.java
nquinquenel: Yes, those are common arguments. I'll try to clean this up with a shared helper maybe

.github/workflows/build.yml
damien-urruty-sonarsource: It seems to be supported: https://mise.jdx.dev/dev-tools/backends/dotnet.html

nquinquenel: I replaced everything with mise, I had to add `install_args: --jobs 1` to [disable](https://mise.jdx.dev/cli/install.html#mise-install) parallel installation, it was causing errors

damien-urruty-sonarsource (approved):

PR #339, opened 2026-07-01
SLOMNI-132 Fetch the build number before other jobs

damien-urruty-sonarsource:
----
## Summary by Gitar

- **CI workflow optimization:**
  - Added `build_number` job to fetch build number centrally before other stages.
  - Updated `build_maven_parent`, `build_dotnet`, `qa_linux`, `qa_windows`, `validate_linux`, `validate_windows`, and `promote` jobs to consume `BUILD_NUMBER` via the new `build_number` dependency.

<sub>This will update automatically on new commits.</sub>

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -18,14 +18,26 @@ permissions:
   contents: write
 
 jobs:
+  build_number:
+    runs-on: github-ubuntu-latest-s
+    name: Get build number
+    outputs:
+      BUILD_NUMBER: ${{ steps.build-number.outputs.BUILD_NUMBER }}
+    permissions:
+      id-token: write
+    steps:
+      - uses: SonarSource/ci-github-actions/get-build-number@657e44543f55422ffd28b91d62fe5386f07d6030 # 1.6.0
+        id: build-number
+
   build_maven_parent:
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -18,14 +18,26 @@ permissions:
   contents: write
 
 jobs:
+  build_number:
+    runs-on: github-ubuntu-latest-s
```
nquinquenel, line 22: I don't remember, didn't we migrate to `sonar-xs` at some point? Maybe not in scope of this PR though

nquinquenel (approved):

.github/workflows/build.yml
damien-urruty-sonarsource: Not for this repo. What we migrated in some other repos is `sonar-XXX-public` to `sonar-XXX`. I just noticed on the xtranet that `github-ubuntu-latest-s` is marked as deprecated but I didn't hear any communication about it, so it should be fine for now

damien-urruty-sonarsource: Asked the question to Eng XP to be sure, I will create follow-up tickets if needed

PR #350, opened 2026-07-17
Fix Unicode path matching

sky-ricardogomes:
## Summary

Fixes the OmniSharp-to-InputFile path correlation for Windows paths containing Unicode characters.

## Root cause

The sensor compared OmniSharp's native Windows path against the engine file representation indirectly. Input files are represented by percent-encoded file URIs, causing diagnostics for accented paths to be dropped before RPC publication.

## Changes

- Match files by converting each `InputFile` URI through `Paths.get(URI)`.
- Normalize absolute paths and Unicode NFC before comparison.
- Add unit and end-to-end coverage for `diretório/TipoPrestaçãoColeção.cs`.

## Validation

- `OmnisharpSensorTests` — 13 tests passed.
- `OmnisharpIntegrationTests#analyzeCSharpFileWithUnicodePath` — passed.
- `mvn -DskipIts -DskipTests package` — passed.

### DIFF
--- a/its/src/test/java/org/sonarsource/sonarlint/omnisharp/its/OmnisharpIntegrationTests.java
+++ b/its/src/test/java/org/sonarsource/sonarlint/omnisharp/its/OmnisharpIntegrationTests.java
@@ -220,6 +220,34 @@ class OmnisharpIntegrationTests {
   }
 
   @Test
+  void analyzeCSharpFileWithUnicodePath(@TempDir Path tmpDir) throws Exception {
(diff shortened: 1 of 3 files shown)


georgii-borovinskikh-sonarsource:
https://github.com/SonarSource/sonarlint-omnisharp/pull/354

sky-ricardogomes:
Thanks for opening #354 and for taking this forward. I’ll follow it and can validate the fix on Rider/IntelliJ using the Unicode-path reproduction case. Thanks again!

